PR #296, opened 2018-11-09
WIP extended k8s tests

hardbyte:
My plan is to run the rather longer integration tests and a benchmark against a "default" rather than minimal deployment on k8s.


hardbyte:
abandoned. Would rather invest in getting testing onto public infrastructure.

Issue #327, opened 2019-02-17
Handle missing results during aggregation step


hardbyte:
See traceback in https://github.com/n1analytics/entity-service/issues/324#issuecomment-464414052

Requires a check before assuming elements of `similarity_result_files` are 2-tuples.

hardbyte:
Was addressed back in Feb.

PR #328, opened 2019-02-17
Handle missing results during aggregation

hardbyte:
Small fix to address #327. 

Adds error handling in the aggregation task.

### DIFF
--- a/backend/entityservice/tasks/base_task.py
+++ b/backend/entityservice/tasks/base_task.py
@@ -79,7 +79,7 @@ class TracedTask(BaseTask):
         except:
             pass
         if parent is None:
-            logger.debug('bugger, parent is None')
+            logger.debug('Unable to trace across tasks as parent span is None')
         with self.tracer.start_span(getattr(self, 'span_name', self.name), child_of=parent) as wrapper_span:
             self._span = wrapper_span
             for span_tag in getattr(self, 'args_as_tags', []):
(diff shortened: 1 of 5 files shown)


backend/entityservice/tasks/comparing.py
```diff
@@ -220,7 +222,13 @@ def aggregate_comparisons(similarity_result_files, project_id, run_id, parent_sp
     files = []
     data_size = 0
 
-    for num, filename in similarity_result_files:
+    for res in similarity_result_files:
+        if res is None:
+            log.warning("Missing results during aggregation. Stopping processing.")
+            return
```
wilko77: You do a silent fail here, as well as a few lines up. This would result in the run status staying 'running' forever. I think that whenever we do not intend to actually compute a result, we should set the status to 'error'...

backend/entityservice/tasks/base_task.py
```diff
@@ -105,7 +105,16 @@ def celery_bug_fix(*args, **kwargs):
...
 @celery.task(base=BaseTask, ignore_result=True)
 def on_chord_error(*args, **kwargs):
-    logger.info("An error occurred while processing the chord's tasks")
+    """
+    Record that a task has encountered an error, mark the run as failed.
+
+    :param args: A 1-tuple starting with the taskid.
+    :param kwargs: Keyord arguments to the task e.g. {'run_id': '...', }
+    """
+    task_id = args[0]
```
wilko77: according to this: https://github.com/celery/celery/issues/3709#issuecomment-407491389 The error task gets called only with a result id which corresponds to the `ChordError` result.
Maybe we can get hold of that result and maybe maybe it contains some useful information....

hardbyte: that comment may be refering to a different version or is at least partially wrong. The error handler's `kwargs` component has the task's arguments - which I use to extract `run_id`. Feel free to look into it separately though.

PR #333, opened 2019-02-27
Use new solver API in anonlink

nbgl:
Old solver is deprecated, so let’s use the new one.


nbgl:
The Kubernetes deployment is failing, but (a) it’s optional and (b) I don’t think it’s caused by this PR, so I’ll merge now.

nbgl:
Thank you @hardbyte for fixing a bug in this PR. Is this the thing that caused the solver test to time out? If it is, then we need to be more explicit about when tasks are failing with errors when testing, as opposed to timing out. If it isn’t then we need to test the solver step properly.

PR #334, opened 2019-03-04
Dependency hell

nbgl:
1. Fix Tornado version. Tornado released v6 on 1st March, which removes some deprecated APIs that we use. This caused things to crash.
2. Update Redis to 3.2.0.


hardbyte:
We use `Tornado`?

nbgl:
`Required-by: threadloop, opentracing-instrumentation, jaeger-client, flower`

Issue #335, opened 2019-03-05
Sort similarity scores


nbgl:
Similarity scores need to be sorted before solving. There are some helper tools in `anonlink.concurrency` for this.

nbgl:
I’ve disabled two tests in `test_result_correctness.py` to account for the output of anonlink v0.11 being correct, but the Entity Service being still incorrect. They need to be reenabled.

hardbyte:
This was incorrectly close by a comment in https://github.com/data61/anonlink-entity-service/pull/339

nbgl:
The greedy solver needs to consider scores from highest to lowest to maximise accuracy. Sorting before solving is the usual way of doing this (I describe a more efficient way in data61/anonlink#212, but it is more complex to implement).

The Entity Service _does_ actually sort similarity scores now, so this issue may be closed. The code is [here](https://github.com/data61/anonlink-entity-service/commit/ffb2c7a8700b3c3945190321886f5145367369ed?diff=unified#diff-30fb0765378c9296147bce2d345daed9R239) as part of #339.

A concern might be that this aggregation is single-threaded, so it might not be the most efficient. Parallelising this merge sort is not a research question, but a software engineering one. (Not hard, but annoying.) But this should be identified as a bottleneck before any work is done, and it should be a separate issue.

hardbyte:
Thanks for the clarification Jakub

PR #336, opened 2019-03-06
Use anonlink’s new similarities api

nbgl:
Use the new Anonlink API for similarity computation. This has three consequences:
1. Anonlink no longer requires popcounts or indices to be passed along with Bloom filters. We thus stop passing them through the Entity Service and we change the binary serialisation of the filters to not include them. This means that **the binary upload API has changed in this PR to be simpler.**
2. Anonlink sorts the similarity scores correctly before solving. Previously we expected the ES to return the same matching as pure Anonlink. This is no longer the case. Previously, both Anonlink and ES were wrong; now Anonlink is no longer wrong so the mappings don’t necessarily match. I’ve disabled two tests in `test_results_correctness.py` because of this; this is captured in #335 and those tests can be reënabled when ES correctly sorts similarity scores. There are some utilities for this in `anonlink.serialization`.
3. The new anonlink API accepts bytes in place of bitarrays. This means that we stop converting hashes to bitarrays.


backend/entityservice/tasks/comparing.py
```diff
@@ -148,10 +148,11 @@ def compute_filter_similarity(chunk_info_dp1, chunk_info_dp2, project_id, run_id
...
     span.log_kv({'size1': chunk_dp1_size, 'size2': chunk_dp2_size})
-    chunk_results = anonlink.entitymatch.calculate_filter_similarity(chunk_dp1, chunk_dp2,
-                                                                     threshold=threshold,
-                                                                     k=min(chunk_dp1_size, chunk_dp2_size),
-                                                                     use_python=False)
+    chunk_results = anonlink.candidate_generation.find_candidate_pairs(
+        (chunk_dp1, chunk_dp2),
+        anonlink.similarities.dice_coefficient_accelerated,
+        threshold,
+        k=min(chunk_dp1_size, chunk_dp2_size))
```
hardbyte: Recalling our conversation last week about `k` restricting the number of similarity scores to the size of the smaller party which is a problem when outputting similarity scores directly. 

Do you see any issue with changing this to `max`? Probably makes more sense to do in a subsequent issue/PR but just wanted to flag it.

backend/entityservice/tasks/encoding_uploading.py
```diff
@@ -39,29 +36,24 @@ def handle_raw_upload(project_id, dp_id, receipt_token, parent_span=None):
...
 
-        log.info(f"Processed {len(clkcounts)} hashes")
+        log.info(f"Processed {i + 1} hashes")
 
     # We peek at the first element as we need the encoding size
     # for the ret of our processing pipeline
-    python_filters = more_itertools.peekable(filter_generator())
-    # Note the len of a bitarray is returned in bits but we require
-    # this to be a multiple of 8 so we use bytes.
-    uploaded_encoding_size = len(python_filters.peek()[0])//8
```
hardbyte: Great to remove this. I think that was the only use of  the`more_intertools` dependency so that could now be removed.

backend/entityservice/tasks/comparing.py
nbgl: The only possible issue would be performance if the two datasets are of very different size. It would change the time complexity from _O_(min(_n_, _m_)²) to _O_(_nm_).

nbgl: Also, I’ve historically been in favour of just getting rid of this _k_ parameter from our code altogether. It’s meant to prevent the similarity computation from taking up too much memory (when there’s too many pairs above the threshold) but we don’t actually use it anywhere. Further, as it’s currently implemented, it’s not symmetric, so it causes you to get different results if you swap the order of the dataset. Also, it was a serious pain to implement maintaining only _k_ closest matches for a record when the similarity computation is distributed (nb: this is implemented in anonlink; still needs to be called from ES).

backend/entityservice/tasks/encoding_uploading.py
nbgl: Removed!

backend/entityservice/serialization.py
```diff
@@ -28,11 +28,8 @@ def list_to_bytes(python_object):
     raise TypeError(repr(python_object) + ' is not valid bytes')
 
 
-def deserialize_bitarray(bytes_data):
-    ba = bitarray(endian='big')
-    data_as_bytes = base64.decodebytes(bytes_data.encode())
-    ba.frombytes(data_as_bytes)
-    return ba
+def deserialize_bytes(bytes_data):
+    return base64.b64decode(bytes_data)
```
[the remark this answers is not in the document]
nbgl: This is a change to the internal representation of a CLK after it’s been deserialised from Base64. It should not be visible to the user.

nbgl:
I’ve checked the docs and they mention nothing about the binary upload format, so there is nothing to update. (I will open a new issue to write the docs.)

The Base64 serialisation format remains unchanged, so there is nothing to update there either.

PR #343, opened 2019-03-19
Add support for external postgresql

hardbyte:
This replaces PR #342 from @Muelsy - for security reasons we don't run tests on our internal infrastructure from external contributors.

> I needed to use an external postgres provider for my test cluster, which meant I had to modify some templates. This PR enables users to connect the entity service to a a postgres server of their choosing. Pretty basic - if/else condition to see if postgres is being provisioned, if it is then set the database server to {{.Release.Name }}-{{ .Values.postgresql.nameOverride}}. If provisioning is disabled, use {{ .Values.postgresql.nameOverride }} instead.
> 
> This is working on my home cluster, and it passed the integration tests.

(@nbgl I've reviewed the changes but I don't want to approve my own PR.)

### DIFF
--- a/deployment/entity-service/templates/configmap.yaml
+++ b/deployment/entity-service/templates/configmap.yaml
@@ -12,7 +12,13 @@ data:
   REDIS_SERVER: {{ .Values.redis.server }}
   REDIS_USE_SENTINEL: {{ .Values.redis.use_sentinel }}
   {{ end }}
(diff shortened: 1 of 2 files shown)


hardbyte:
Sorry @nbgl  it looks like the extra rubber stamping wasn't required - the other PR is now getting the status checks and has my review. Thanks though :-)

PR #348, opened 2019-04-01
Update postgres

hardbyte:
Update to new postgres chart which has global configurations as documented [here](https://github.com/helm/charts/tree/master/stable/postgresql#use-of-global-variables)

### DIFF
--- a/Jenkinsfile.groovy
+++ b/Jenkinsfile.groovy
@@ -281,7 +281,7 @@ node('helm && kubectl') {
                 helm upgrade --install --wait --namespace ${NAMESPACE} ${DEPLOYMENT} . \
                     -f values.yaml -f minimal-values.yaml -f test-versions.yaml \
                     --set api.app.debug=true \
-                    --set postgresql.postgresqlPassword=notaproductionpassword \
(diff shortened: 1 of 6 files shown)


Jenkinsfile.groovy
```diff
@@ -281,7 +281,7 @@ node('helm && kubectl') {
                 helm upgrade --install --wait --namespace ${NAMESPACE} ${DEPLOYMENT} . \
                     -f values.yaml -f minimal-values.yaml -f test-versions.yaml \
                     --set api.app.debug=true \
-                    --set postgresql.postgresqlPassword=notaproductionpassword \
+                    --set global.postgresql.postgresqlPassword=notaproductionpassword \
```
nbgl: Why the `global`? Does it have a meaning or is it just a cosmetic change?

hardbyte: Bit of both. In our case it doesn't (yet) make any difference but the `global` does have significance - in that when you have multiple sub-charts that want to access the same variable it needs to be declared as global. https://github.com/helm/helm/blob/master/docs/chart_template_guide/subcharts_and_globals.md#global-chart-values

Issue #323, opened 2019-02-04
Configure logging using k8s configmap


hardbyte:
Expose standard way of configuring logging via json and/or yaml. In essence we should support `Iogging.config.dictConfig`, and at deployment the user can provide the configuration content. 

Want to allow users to be able to define their own handlers and their respective log levels and document how/where. 

Also see [protari logging docs](https://docs.protari.d61.io/dev/setting-up/logging/).

hardbyte:
Last thing required before closing this is configuration of logging in a k8s configmap.

gusmith:
Will be closed by #448

gusmith:
#448 merged in.

PR #349, opened 2019-04-06
Configurable logging

hardbyte:
Instead of having the logging config all hardcoded we can read from a yaml file.

Addresses issue #323.

Fixes a few issues with logging too (e.g. including the timestamp twice).

### DIFF
--- a/backend/entityservice/async_worker.py
+++ b/backend/entityservice/async_worker.py
@@ -5,7 +5,7 @@ from celery.signals import task_prerun
 import structlog
 
 from entityservice.settings import Config
-
+from entityservice.logger_setup import setup_structlog
 
 celery = Celery('tasks',
                 broker=Config.CELERY_BROKER_URL,
@@ -21,31 +21,15 @@ celery.conf.worker_prefetch_multiplier = Config.CELERYD_PREFETCH_MULTIPLIER
 celery.conf.worker_max_tasks_per_child = Config.CELERYD_MAX_TASKS_PER_CHILD
 
 
-structlog.configure(
-    processors=[
-        structlog.stdlib.add_logger_name,
-        structlog.stdlib.add_log_level,
-        structlog.stdlib.PositionalArgumentsFormatter(),
-        structlog.processors.StackInfoRenderer(),
-        structlog.processors.format_exc_info,
-        structlog.dev.ConsoleRenderer()
-    ],
-    context_class=structlog.threadlocal.wrap_dict(dict),
-    logger_factory=structlog.stdlib.LoggerFactory(),
-)
-
-# Set logging level of other python libraries that we use
-logging.getLogger('celery').setLevel(logging.WARNING)
-logging.getLogger('jaeger_tracing').setLevel(logging.WARNING)
-
 # Set up our logging
-logger = structlog.wrap_logger(logging.getLogger('celery.es'))
+setup_structlog()
+logger = structlog.wrap_logger(logging.getLogger('entityservice.tasks'))
 if Config.DEBUG:
-    logging.getLogger('celery.es').setLevel(logging.DEBUG)
-    logging.getLogger('celery').setLevel(logging.INFO)
+    logging.getLogger('entityservice.tasks').setLevel(logging.DEBUG)
+    #logging.getLogger('celery').setLevel(logging.INFO)
 
 
-logger.info("Setting up celery worker")
+logger.info("Setting up entityservice worker")
 logger.debug("Debug logging enabled")
 
 
--- a/backend/entityservice/async_worker.py
+++ b/backend/entityservice/async_worker.py
@@ -24,11 +24,6 @@ celery.conf.worker_max_tasks_per_child = Config.CELERYD_MAX_TASKS_PER_CHILD
 # Set up our logging
 setup_structlog()
 logger = structlog.wrap_logger(logging.getLogger('entityservice.tasks'))
-if Config.DEBUG:
-    logging.getLogger('entityservice.tasks').setLevel(logging.DEBUG)
-    #logging.getLogger('celery').setLevel(logging.INFO)
-
-
 logger.info("Setting up entityservice worker")
 logger.debug("Debug logging enabled")
 
--- a/backend/entityservice/logger_setup.py
+++ b/backend/entityservice/logger_setup.py
@@ -1,28 +1,26 @@
-import os
 import logging.config
+import os
+from pathlib import Path
 import structlog
-
 import yaml
 
 
 def setup_logging(
-    default_path='logging.yaml',
+    default_path='default_logging.yaml',
     default_level='INFO',
     env_key='LOG_CFG'
 ):
     """
     Setup logging configuration
     """
-    path = os.path.dirname(os.path.abspath(__file__)) + '/' + default_path
-    value = os.getenv(env_key, None)
-    if value:
-        path = value
-    if os.path.exists(path):
+    path = os.getenv(env_key, Path('.') / default_path)
+
+    try:
         with open(path, 'rt') as f:
             config = yaml.safe_load(f.read())
         logging.config.dictConfig(config)
-    else:
-        print("YAML file doesn't exist")
+    except FileNotFoundError:
+        print("Logging config YAML file doesn't exist. Falling back to defaults")
         default_level = _str_to_level(default_level)
         logging.basicConfig(level=default_level)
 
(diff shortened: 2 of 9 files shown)


backend/entityservice/logger_setup.py
```diff
@@ -0,0 +1,90 @@
...
+
+def setup_logging(
+    default_path='logging.yaml',
+    default_level='INFO',
+    env_key='LOG_CFG'
+):
+    """
+    Setup logging configuration
+    """
+    path = os.path.dirname(os.path.abspath(__file__)) + '/' + default_path
```
wilko77: how about `os.path.join` instead?

backend/entityservice/async_worker.py
```diff
@@ -21,31 +21,15 @@
...
-logging.getLogger('jaeger_tracing').setLevel(logging.WARNING)
-
 # Set up our logging
-logger = structlog.wrap_logger(logging.getLogger('celery.es'))
+setup_structlog()
+logger = structlog.wrap_logger(logging.getLogger('entityservice.tasks'))
 if Config.DEBUG:
-    logging.getLogger('celery.es').setLevel(logging.DEBUG)
-    logging.getLogger('celery').setLevel(logging.INFO)
+    logging.getLogger('entityservice.tasks').setLevel(logging.DEBUG)
```
wilko77: you also define a log level in the logging.yaml file. How does that relate to `Config.DEBUG`? And which value is set eventually?

backend/entityservice/logger_setup.py
```diff
@@ -0,0 +1,90 @@
...
+    def filter(self, record):
+        return 1 if record.levelno < logging.ERROR else 0
+
+
+def _str_to_level(lvl):
+    """
+    Convenience function to convert a log level string to the numeric rep
+    """
+    lvl = lvl.strip().lower()
+    if lvl == 'warn':
```
nbgl: Python’s logging nomenclature prefers ‘warning’ to ‘warn’.

backend/entityservice/logger_setup.py
```diff
@@ -0,0 +1,90 @@
...
+    elif lvl == 'info':
+        return logging.INFO
+    elif lvl == 'debug':
+        return logging.DEBUG
+    elif lvl == 'critical':
+        return logging.CRITICAL
+    elif lvl == 'error':
+        return logging.ERROR
+    else:
+        return logging.INFO
```
nbgl: This fails silently if `lvl` is unexpected. Instead, the `else` clause should raise an exception.

backend/entityservice/logger_setup.py
```diff
@@ -0,0 +1,90 @@
...
+    if lvl == 'warn':
+        return logging.WARNING
+    elif lvl == 'info':
+        return logging.INFO
+    elif lvl == 'debug':
+        return logging.DEBUG
+    elif lvl == 'critical':
+        return logging.CRITICAL
+    elif lvl == 'error':
+        return logging.ERROR
```
nbgl: Nit: A more Pythonic way to implement this would be with a dictionary lookup:
```python
lvl = lvl.strip().lower()
return {'info': logging.INFO, …}[lvl]
```

backend/entityservice/logger_setup.py
```diff
@@ -0,0 +1,90 @@
...
+    )
+
+
+class StdErrFilter(logging.Filter):
+    """
+    Filter for the stderr stream
+    Doesn't print records below ERROR to stderr to avoid dupes
+    """
+    def filter(self, record):
+        return 0 if record.levelno < logging.ERROR else 1
```
nbgl: I suspect `return record.levelno >= logging.ERROR` would suffice. After all, True == 1 and False == 0.

backend/entityservice/logger_setup.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+class StdOutFilter(logging.Filter):
+    """
+    Filter for the stdout stream
+    Doesn't print records above WARNING to stdout to avoid dupes
+    """
+    def filter(self, record):
+        return 1 if record.levelno < logging.ERROR else 0
```
nbgl: Nit: as above

backend/entityservice/logger_setup.py
nbgl: Use `pathlib` or at least `os.path`

backend/entityservice/logger_setup.py
```diff
@@ -0,0 +1,90 @@
...
+    default_level='INFO',
+    env_key='LOG_CFG'
+):
+    """
+    Setup logging configuration
+    """
+    path = os.path.dirname(os.path.abspath(__file__)) + '/' + default_path
+    value = os.getenv(env_key, None)
+    if value:
+        path = value
```
nbgl: `path = os.getenv(env_key, path)` can replace those three lines

backend/entityservice/logger_setup.py
```diff
@@ -0,0 +1,90 @@
...
+    if os.path.exists(path):
+        with open(path, 'rt') as f:
+            config = yaml.safe_load(f.read())
+        logging.config.dictConfig(config)
+    else:
+        print("YAML file doesn't exist")
```
nbgl: This will be hard to decipher amongst many lines of output.

```suggestion
        print(“Logging configuration YAML file doesn't exist. Falling back to defaults.")
```

backend/entityservice/logger_setup.py
hardbyte: I'll raise you a [pathlib](https://docs.python.org/3/library/pathlib.html#module-pathlib)

backend/entityservice/async_worker.py
hardbyte: `Config.DEBUG` is an old setting that toggles between default and verbose logging based on an environment variable. 

I think I should remove it entirely as the same could be achieved by shipping with a `verbose_logging.yaml` and a `default_logging.yaml`

backend/entityservice/logger_setup.py
```diff
@@ -0,0 +1,90 @@
...
+    env_key='LOG_CFG'
+):
+    """
+    Setup logging configuration
+    """
+    path = os.path.dirname(os.path.abspath(__file__)) + '/' + default_path
+    value = os.getenv(env_key, None)
+    if value:
+        path = value
+    if os.path.exists(path):
```
nbgl: Instead of seeing if the path exists, try opening the file and catch the exception.

```python
try:
    f = open(path)
except FileNotFoundError:
    # handle
else:
    with f:
        # read file
````

File: backend/entityservice/logger_setup.py [hunk 11]
```diff
@@ -0,0 +1,90 @@
...
+            structlog.stdlib.add_log_level,
+            structlog.stdlib.PositionalArgumentsFormatter(),
+            structlog.processors.TimeStamper(fmt='iso'),
+            structlog.processors.StackInfoRenderer(),
+            structlog.processors.format_exc_info,
+
+            # Uncomment ONE Renderer:
+            #structlog.processors.KeyValueRenderer(),
+            #structlog.processors.JSONRenderer(),
+            structlog.dev.ConsoleRenderer()
```
nbgl: Would it make more sense to have this a settable via the config file?

hardbyte: Yeah I think making all of structlog configurable would be a good extension - I'll create an issue rather than add it now though.

backend/entityservice/logger_setup.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+    Filter for the stderr stream
+    Doesn't print records below ERROR to stderr to avoid dupes
+    """
+    def filter(self, record):
+        return record.levelno >= logging.ERROR
+
+
+StdErrFilter = ErrorLevelFilter
+StdOutFilter = ErrorLevelFilter
```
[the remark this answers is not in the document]
hardbyte: Uhh good catch, I just saw what looked like entirely duplicate code and missed that

backend/entityservice/logger_setup.py
```diff
@@ -0,0 +1,92 @@
...
+):
+    """
+    Setup logging configuration
+    """
+    path = os.getenv(env_key, Path(__file__).parent / default_path)
+    try:
+        with open(path, 'rt') as f:
+            config = yaml.safe_load(f)
+        logging.config.dictConfig(config)
+        msg = "Loaded logging config from file"
```
nbgl: It might make more sense to move this string literal to [line 32](https://github.com/n1analytics/entity-service/pull/349/files#diff-a6e1643a794254ae9581fb40cffdf0dcR32) where it is actually used.

PR #352, opened 2019-04-10
Use Anonlink chunking API

nbgl:
As part of this, we remove `chunk_size` from `runs` in the database as the chunk size no longer depends on the job size.

### DIFF
--- a/backend/entityservice/settings.py
+++ b/backend/entityservice/settings.py
@@ -64,13 +64,7 @@ class Config(object):
 
     # Number of comparisons per chunk. Default is to interpolate between the min
     # of 100M and the max size of 1B based on the JOB_SIZE parameters.
-    SMALL_COMPARISON_CHUNK_SIZE = int(os.getenv('SMALL_COMPARISON_CHUNK_SIZE', '100000000'))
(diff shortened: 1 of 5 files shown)


backend/entityservice/settings.py
```diff
@@ -64,13 +64,7 @@ class Config(object):
...
     # Number of comparisons per chunk. Default is to interpolate between the min
     # of 100M and the max size of 1B based on the JOB_SIZE parameters.
-    SMALL_COMPARISON_CHUNK_SIZE = int(os.getenv('SMALL_COMPARISON_CHUNK_SIZE', '100000000'))
-    LARGE_COMPARISON_CHUNK_SIZE = int(os.getenv('LARGE_COMPARISON_CHUNK_SIZE', '1000000000'))
-
-    # Anything smaller that this will use SMALL_COMPARISON_CHUNK_SIZE
-    SMALL_JOB_SIZE = int(os.getenv('SMALL_JOB_SIZE', '100000000'))
-    # Anything greater than this will use LARGE_COMPARISON_CHUNK_SIZE
-    LARGE_JOB_SIZE = int(os.getenv('LARGE_JOB_SIZE', '100000000000'))
+    CHUNK_SIZE_AIM = int(os.getenv('CHUNK_SIZE_AIM', '300000000'))
```
hardbyte: The comment above needs updating.

nbgl: Thanks!

PR #358, opened 2019-04-24
 Generalise worker to many parties

nbgl:
This generalises the worker functions to many parties, to make it easier to enable multi-party linkage. Changes to tests are not necessary as everything is equivalent for 2-party linkage.

### DIFF
--- a/backend/entityservice/tasks/comparing.py
+++ b/backend/entityservice/tasks/comparing.py
@@ -36,7 +36,8 @@ def create_comparison_jobs(project_id, run_id, parent_span=None):
 
         dp_ids = get_dataprovider_ids(conn, project_id)
         assert len(dp_ids) >= 2, "Expected at least 2 data providers"
-        log.info("Starting comparison of CLKs from data provider ids: {}, {}".format(dp_ids[0], dp_ids[1]))
+        log.info(f"Starting comparison of CLKs from data provider ids: "
+                 f"{', '.join(map(str, dp_ids))}")
         current_span = create_comparison_jobs.span
 
         if not check_project_exists(conn, project_id) or not check_run_exists(conn, project_id, run_id):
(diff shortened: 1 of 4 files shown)


backend/entityservice/tasks/comparing.py
```diff
@@ -152,20 +151,8 @@ def compute_filter_similarity(chunk_info, project_id, run_id, threshold, encodin
 
     t4 = time.time()
 
-    partial_sparse_result = []
-    # offset chunk's index
-    offset_0, _ = chunk_info_dp1['range']
-    offset_1, _ = chunk_info_dp2['range']
```
nbgl: Offsetting is dealt with by `anonlink.concurrency.process_chunk` so we don’t need this extra step.

backend/entityservice/tasks/comparing.py
```diff
@@ -186,17 +173,16 @@ def compute_filter_similarity(chunk_info, project_id, run_id, threshold, encodin
...
+    t5 = time.time()
 
     log.info("run={} Comparisons: {}, Links above threshold: {}".format(run_id, comparisons_computed, len(chunk_results)))
-    log.info("Prep: {:.3f} + {:.3f}, Solve: {:.3f}, Progress: {:.3f}, Offset: {:.3f}, Save: {:.3f}, Total: {:.3f}".format(
+    log.info("Prep: {:.3f} + {:.3f}, Solve: {:.3f}, Progress: {:.3f}, Save: {:.3f}, Total: {:.3f}".format(
         t1 - t0,
         t2 - t1,
         t3 - t2,
         t4 - t3,
-        t4 - t4,
```
nbgl: This line was a bug. Should have said `t5 - t4` the whole time (instead of `t4 - t4`).

backend/entityservice/tasks/comparing.py
```diff
@@ -52,33 +53,30 @@ def create_comparison_jobs(project_id, run_id, parent_span=None):
...
     for chunk_info in chunk_infos:
         chunk_dp1_info, chunk_dp2_info = chunk_info
-        assert chunk_dp1_info['datasetIndex'] == 0
-        chunk_dp1_info['storeFilename'] = filters1_object_filename
-        assert chunk_dp2_info['datasetIndex'] == 1
-        chunk_dp2_info['storeFilename'] = filters2_object_filename
+        chunk_dp1_info['storeFilename'] \
+            = filters_object_filenames[chunk_dp1_info['datasetIndex']]
+        chunk_dp2_info['storeFilename'] \
+            = filters_object_filenames[chunk_dp2_info['datasetIndex']]
```
hardbyte: To avoid confusion in the future these shouldn't be called `chunk_dp1...` and `chunk_dp2...` anymore.

nit: not a fan of splitting simple assignments over two lines just to stay under a max line length.

backend/entityservice/tasks/comparing.py
hardbyte: Nothing sparks joy like deleting code that doesn't spark joy

nbgl: Yes, I was very happy about this!

backend/entityservice/tasks/comparing.py
nbgl: I’ve changed it to something else. Please let me know if you think it’s better.

PR #359, opened 2019-04-25
Fix chunk merging (and one other minor thing)

nbgl:
Two bugfixes:

1. The merging of similarities is currently crashing because of errors that were not caught by the tests. This is bad. These two line changes fix it.

2. Read `number_parties` correctly when making a project.


nbgl:
I’m unsure and confused. It makes sense that the unit tests would not be trying massive problems (and hence there wouldn’t be multiple chunks), but I would have thought that Jenkins runs big ones. There is alternative possibility that the chunking code is not working correctly. I’ve made issue #360 to track this.

PR #362, opened 2019-04-28
Have jenkins fail if the benchmark errors

### LINKED ISSUE #361: jenkins didn't report failed benchmark
benchmark errored https://magic-ci-atp.it.csiro.au/job/entity-service/job/develop/108/execution/node/77/log/ but jenkins gave it the thumbs up anyway.

Is that the desired behavior?

hardbyte:
- Fails if the benchmark container exits incorrectly (due to bad configuration).
- Fails if the benchmark results contain status=ERROR entries.
- Removes the default server address for benchmark container.

Closes #361

### DIFF
--- a/Jenkinsfile.groovy
+++ b/Jenkinsfile.groovy
@@ -165,7 +165,7 @@ node('docker&&multicore&&ram') {
           def benchmarkResults = readJSON file: 'results.json'
           print(benchmarkResults["version"])
           for (String item : benchmarkResults["experiments"]) {
-               if(item["status"] == "ERROR") {
+               if (item["status"] != "completed") {
                  throw new Exception("Benchmark failed\n" + item["description"])
                }
           }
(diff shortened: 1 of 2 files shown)


Jenkinsfile.groovy
```diff
@@ -149,11 +149,27 @@ node('docker&&multicore&&ram') {
...
+          sh """
           docker cp ${benchmarkContainerName}:/app/results.json results.json
           """
-
           archiveArtifacts artifacts: "results.json", fingerprint: true
+          // Read the JSON file and fail if there was an ERROR
+          def benchmarkResults = readJSON file: 'results.json'
+          print(benchmarkResults["version"])
+          for (String item : benchmarkResults["experiments"]) {
+               if(item["status"] == "ERROR") {
```
wilko77: I would test for success and otherwise fail. (There might be other statuses like 'running', etc.)
Oh, and add a space after the `if`. 
```if (item["status"] != "completed") {```

hardbyte: fair enough, I'm going to trust you on the `"completed"`

PR #363, opened 2019-04-28
Have Jenkins test notebooks

### LINKED ISSUE #110: Automatically test/run tutorial notebooks
write tests for the correct workings of the tutorial  notebooks

hardbyte:
Uses the `nbval` plugin for pytest, runs inside a docker container on Jenkins.

Like clkhash only tests that none of the notebook cells error.

Closes #110

### DIFF
--- a/Jenkinsfile.groovy
+++ b/Jenkinsfile.groovy
@@ -103,10 +103,10 @@ node('docker&&multicore&&ram') {
             quay.io/n1analytics/entity-docs-tutorial:latest
         """
         tutorialTestingcontainer.watchLogs()
-        gitCommit.setSuccessStatus(gitContextLocalBenchmark)
+        gitCommit.setSuccessStatus(gitContextTutorialTests)
       } catch (err) {
         print("Error in testing tutorial notebooks:\n" + err)
-        gitCommit.setFailStatus(gitContextLocalBenchmark)
+        gitCommit.setFailStatus(gitContextTutorialTests)
         throw err
       }
     }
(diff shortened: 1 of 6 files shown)


Jenkinsfile.groovy
```diff
@@ -73,7 +77,8 @@ node('docker&&multicore&&ram') {
       try {
         echo("Start all the containers (including tests)")
         sh "docker-compose -v"
-          composeCmd(composeProject, "up -d db minio redis backend db_init worker nginx tests")
+        composeCmd(composeProject, "up -d db minio redis backend db_init worker nginx")
+        sleep 20
```
wilko77: what are those sleep commands for? And is 20 the right value?

docs/tutorial/Similarity Scores.ipynb
```diff
@@ -813,38 +893,69 @@
...
    ],
    "source": [
     "look_at_per_field_accuracy(threshold = 0.95, num_samples = 100)"
    ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 26,
+   "metadata": {},
+   "outputs": [
```
wilko77: what's that output for? There is no source for it...

Jenkinsfile.groovy
hardbyte: I lowered and moved the 30 second sleep from line 90 (before following the output of the integration test container). I suspect it gives a bit of buffer for pulling docker images.

Jenkinsfile.groovy
```diff
@@ -82,12 +87,29 @@ node('docker&&multicore&&ram') {
       }
     }
 
+    stage('Tutorial Tests') {
+      //gitCommit.setInProgressStatus(gitContextDocumentation);
```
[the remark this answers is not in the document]
hardbyte: Will do!

docs/tutorial/Dockerfile
```diff
@@ -0,0 +1,15 @@
+# Note we expect the user to mount the sources to build...
```
hardbyte: this is incorrect. I'll remove....

PR #366, opened 2019-05-07
Test project API for multiparty support

nbgl:


backend/entityservice/tests/test_project.py
```diff
@@ -1,52 +1,74 @@
...
 from entityservice.tests.util import generate_overlapping_clk_data, get_project_description, delete_project
 
 
-def _check_new_project_response_fields(new_project_data):
+@pytest.fixture(scope='module',
+                params=[2, 3, 4, None])
+def number_parties(request):
+    n = request.param
+    # None is what we use to test handling of default values
+    number_parties_dict = {} if n is None else {'number_parties': n}
```
nbgl: Check handling of default value as well as of specified values.

backend/entityservice/tests/util.py
```diff
@@ -49,15 +50,68 @@ def generate_json_serialized_clks(count, size=128):
...
+    return itertools.chain.from_iterable(
+        itertools.combinations(s, r) for r in range(1, len(s)+1))
 
-    overlap_to = math.floor(min(dataset_sizes)*overlap)
-    for ds in datasets[1:]:
-        ds[:overlap_to] = datasets[0][:overlap_to]
-        random.shuffle(ds)
+
+def generate_overlapping_clk_data(
+    dataset_sizes, overlap=0.9, encoding_size=128, seed=666):
```
nbgl: Generalise existing function to many parties.

backend/entityservice/tests/test_project_with_runs_deletes.py
```diff
@@ -3,6 +3,9 @@
 
 
 def test_delete_project_after_creating_run_with_clks(requests, result_type):
-    project, dp1, dp2 = create_project_upload_fake_data(requests, [100, 100], overlap=0.5, result_type=result_type)
+    project, (dp1, dp2) = create_project_upload_fake_data(
+        requests, [100, 100], overlap=0.5, result_type=result_type)
+    project, (dp1, dp2) = create_project_upload_fake_data(
```
wilko77: why do you repeat that line now? It wasn't there before.
Also, why not `project, _ = ...` like in the other tests? `dp1` and `dp2` are unused.

backend/entityservice/tests/test_project_run_description.py
```diff
@@ -16,11 +16,15 @@ def test_run_description_no_data(requests, project):
...
+
+    project, _ = create_project_upload_fake_data(
+        requests, [100] * number_parties, overlap=0.5, result_type=result_type)
+    run_id = post_run(requests, project, THRESHOLD)
     run = get_run(requests, project, run_id)
 
     assert 'run_id' in run
-    assert 'notes' in run
-    assert 'threshold' in run
+    assert run['notes'] == ''
```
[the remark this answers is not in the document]
nbgl: Sure. I'd rather revert to not checking the value of `notes`, rather than to change `post_run`, because I feel that'd be out of scope for this PR.

backend/entityservice/tests/test_project_with_runs_deletes.py
nbgl: The line repeat is unintentional (copy-paste error?). Thanks for pointing it out!

The unused `dp1` and `dp2` were there before. To notice it I'd have to read the whole function and I didn't need to do that to make this simple typing change. But I've fixed it also.

PR #371, opened 2019-05-23
Fail on invalid result_type/number_parties combination

nbgl:
The only result type that works with multi-party linkage is `”groups”`. This PR makes a project POST request fail if `number_parties` > 2 but `result_type` ≠ groups.

### DIFF
--- a/backend/entityservice/tests/conftest.py
+++ b/backend/entityservice/tests/conftest.py
@@ -175,6 +175,8 @@ def groups_project(request, requests):
 
 
 @pytest.fixture(
-    params=itertools.product(PROJECT_RESULT_TYPES_2P, [3, 4, 5]))
+    params=itertools.chain(
+        itertools.product(PROJECT_RESULT_TYPES_2P, [1, 3, 4, 5]),
+        [(t, 1) for t in PROJECT_RESULT_TYPES_NP]))
 def invalid_result_type_number_parties(request):
     yield request.param
--- a/backend/entityservice/tests/conftest.py
+++ b/backend/entityservice/tests/conftest.py
@@ -127,6 +127,15 @@ def result_type_number_parties(request):
     yield request.param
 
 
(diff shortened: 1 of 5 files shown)


backend/entityservice/tests/test_project.py
```diff
@@ -26,17 +26,16 @@ def _check_new_project_response_fields(new_project_data,
     assert len(new_project_data['update_tokens']) == expected_number_parties
 
 
-def test_simple_create_project(requests, number_parties_or_none):
+def test_simple_create_project(requests, valid_project_param):
```
nbgl: A bunch of these tests are also now parametrised to check output types other than `"mapping"`.

backend/entityservice/views/project.py
```diff
@@ -205,10 +205,11 @@ def authorise_get_request(project_id):
     with DBConn() as dbinstance:
         project_object = db.get_project(dbinstance, project_id)
     logger.info("Checking credentials")
-    if project_object['result_type'] == 'mapping' or project_object['result_type'] == 'similarity_scores':
+    result_type = project_object['result_type']
+    if result_type in {'mapping', 'similarity_scores', 'groups'}:
```
nbgl: Small bugfix.

backend/entityservice/tests/conftest.py
```diff
@@ -163,3 +172,9 @@ def groups_project(request, requests):
     prj = create_project_response(requests, size, overlap, 'groups', encoding_size)
     yield prj
     delete_project(requests, prj)
+
+
+@pytest.fixture(
+    params=itertools.product(PROJECT_RESULT_TYPES_2P, [3, 4, 5]))
```
wilko77: there are other invalid combinations:
`PROJECT_RESULT_TYPES_2P` and 1, `PROJECT_RESULT_TYPES_NP` and 1.

PR #379, opened 2019-06-03
Example: multiparty linkage in ES

nbgl:
I suspect this needs a bit more context. I welcome suggestions as to what that should be.


example/multiparty-example/multiparty-linkage-in-entity-service.ipynb
```diff
@@ -0,0 +1,1429 @@
...
+    "    for i, j in group:\n",
+    "        print(i, datasets[i][j])\n",
+    "    print()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This matching appears impressively accurate for such noisy data. However, there is still clearly some room for improvement. We may be able to improve on this results by fine-tuning the hashing schema or by changing the threshold."
```
wilko77: The user doesn't know how noisy this data is, so (s)he might not be that impressed. Also, `impressively accurate` is astonishingly vague... You could quantify the accuracy.
For the improvement part, why don't you report the results for other thresholds as well?
Maybe point to an artifact where we do some sort of fine-tuning of the hashing schema.

wilko77:
I addressed everything I complaint about. Unfortunately, testing.es... is not working properly, so I run a local service to populate the notebook. We should probably regenerate the notebook outputs once the testing.es... deployment works again.

PR #382, opened 2019-06-06
Remove quay.io pull secret

hardbyte:
To be clear this was a "readonly" secret for a public docker image hosted on quay.io


hardbyte:
Sure can

Issue #289, opened 2018-10-31
Share database connections in celery workers


hardbyte:
It takes some work to connect to the postgresql database and almost every worker task does that multiple times - usually just relying on Python's garbage collection for closing connections. Instead we might manage that state in a per worker pool, or since celery itself maintains a pool of worker processes, we could just init a connection along with each worker. 

```
db_conn = None

@worker_process_init.connect
def init_worker(**kwargs):
    global db_conn
    logger.info('Initializing database connection for worker.')
    db_conn = connect_db()


@worker_process_shutdown.connect
def shutdown_worker(**kwargs):
    global db_conn
    if db_conn:
        logger.info('Closing database connection for worker.')
        db_conn.close()

```

- https://stackoverflow.com/questions/14526249/celery-worker-database-connection-pooling
- https://code.i-harness.com/en/q/2263d77

hardbyte:
Possible bug sighting in the wild during a 1M x 10M match:


```
 2019-06-05T05:58:05.971235Z [debug    ] Connecting to redis            [entityservice.cache] 
2019-06-05T05:58:05.971634Z [debug    ] Looking up read only redis slave using sentinel protocol [entityservice.cache] 
2019-06-05T05:58:05.985530Z [warning  ] fallback connecting to default postgres db [entityservice.database.util] database=postgres host=june-testing-anonlink-db user=postgres
2019-06-05T05:58:06.012383Z [warning  ] Can't connect to database      [entityservice.database.util] 
[2019-06-05 05:58:06 +0000] [9] [ERROR] Socket error processing request.
Traceback (most recent call last):
  File "/var/www/entityservice/database/util.py", line 38, in connect_db
    conn = psycopg2.connect(database=db, user=user, password=pw, host=host)
  File "/usr/lib/python3.6/site-packages/psycopg2/__init__.py", line 130, in connect
    conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
psycopg2.OperationalError: FATAL:  sorry, too many clients already
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/var/www/entityservice/database/util.py", line 41, in connect_db
    conn = psycopg2.connect(database='postgres', user=user, password=pw, host=host)
  File "/usr/lib/python3.6/site-packages/psycopg2/__init__.py", line 130, in connect
    conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
[...]

PR #386, opened 2019-06-12
Tweak database connection tear down and logging

hardbyte:
I was [reading up on the flask Application Context](
http://flask.pocoo.org/docs/1.0/appcontext/#storing-data) and saw that the recommended approach for keeping a single database connection is slightly different to what we have.

I also added a debug log message to this teardown, and tweaked the *verbose logging* config so a developer can easily see the debug message. Note this changes the default docker-compose deployment to make including verbose logging a simple uncomment operation.

See also https://github.com/data61/anonlink-entity-service/issues/385

### DIFF
--- a/backend/entityservice/views/project.py
+++ b/backend/entityservice/views/project.py
@@ -41,12 +41,13 @@ def projects_post(project):
         project_model = models.Project.from_json(project)
         logger = logger.bind(pid=project_model.project_id)
     except models.InvalidProjectParametersException as e:
+        logger.info(f"Denied request to add a new project - {e.msg}", project=project)
         safe_fail_request(400, message=e.msg)
 
     # Persist the new project
     logger.info("Adding new project to database")
     try:
-        with DBConn() as conn:
+        with DBConn(get_db()) as conn:
             project_model.save(conn)
     except Exception as e:
         logger.warn(e)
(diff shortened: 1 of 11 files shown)


hardbyte:
Thanks Guillaume - all fair questions. I'll use this as an opportunity to try clean this up further. The confusion isn't helped that three different containers/micro-services use this `database/util.py`.

- `init_db` is used from flask once per deployment to initialize the database (creating tables etc) - but note this is not run from the `backend` flask app so there would be no benefit to caching a connection on `g`. I'll just leave this one alone except adding a small note to document it better.

- Only the flask backend has an application context `g`. `get_db` is therefore supposed to be the way that the flask app accesses the database to take advantage of an already open connection. However as you note the `DBConn` is often used which creates a new connection - I'll look at addressing this in this PR.

- The celery workers were not trying to cache connections yet, although I've been thinking about adding that later. I won't address that here but I'll check if there is an open issue and open one if not.

In theory `DBConn` is made to be useful for containing database transactions for the two primary microservices - the flask backend and the celery workers. But I didn't include a mechanism for injecting an existing connection - that could also be changed as part of properly addressing this.

Again thanks for digging into the detail.

hardbyte:
@gusmith I've tightened up how we use a database connection from the flask app in this PR. I may enchance it further by introducing pooled connections in #385. 

I'll be tracking the database connection use within the celery workers in #289.

backend/entityservice/views/project.py
```diff
@@ -294,7 +295,7 @@ def upload_json_clk_data(dp_id, clk_json, parent_span):
         )
 
     with opentracing.tracer.start_span('update-db', child_of=parent_span) as span:
```
gusmith: As you're removing some none-used parameter, `span` does not seem used here neither.
Also, the span which is started in named `update-db`, but line 248, there is another span named `update-database` (observed only because there was changes around these lines in this PR :) ). 
Should the naming be consistent?

PR #391, opened 2019-07-01
Release with Multiparty Examples

hardbyte:
Merge recent changes from develop into master.


hardbyte:
Good point - will merge after https://github.com/data61/anonlink-entity-service/pull/394

PR #388, opened 2019-06-24
Feature azure pipeline

gusmith:
Add a azure pipeline build scripts, pointing to https://dev.azure.com/data61/Anonlink
This is currently not building because we need to create a CI user in dockerhub who could push images to `anonlink-app`, `anonlink-ngninx` and `anonlink-benchmark`. The remaining should work now (tested on a project under my name).

It should be doing the same steps are what is done in Jenkins, but some steps are not done:
 - [x] building and publishing the documentation -> use a documentation builder image to allow the frontend to build and serve the documentation.
 - [x] building the docs tutorials and testing them ->  image stored in dockerhub repo `data61/anonlink-docs-tutorials`
 - [x]  remove the jenkinsfile when we are happy with the pipeline in Azure, and change the brnach protection to use the corresponding checks.

Note: some changes may break the jenkins pipeline.

PR #392, opened 2019-07-03
Feature pipeline integration tests

gusmith:
That was quite painful, but it is finally working end to end!!!
The integration step using docker compose has been removed to run these tests only using kubernetes.
We were otherwise running twice the same tests on different deployment.
With this pipeline, we are still testing the docker compose and the kubernetes deployment, but running different tests on each.
On the docker compose deployment, we run the benchmark and the tutorial tests.
On k8s, we run the integration tests. 
It was preferred as the integration tests are BIG, and takes a long time, so it is better t use some beefy infrastructure instead of a single small VM running the whole service (which was tricky because of timeouts).

Note: this is **not** deploying a release, which is out of scope for this work.
Note 2: this is to update the PR #388 removing the integration tests and running them in k8s only.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -41,12 +41,12 @@ stages:
     parameters:
       folder: '.'
       dockerFilePath: './frontend/Dockerfile'
-      imageName: anonlink-nginx
+      imageName: data61/anonlink-nginx
       jobName: 'anonlink_nginx'
   - template: .azurePipeline/templateDockerBuildPush.yml
     parameters:
       folder: './backend'
-      imageName: anonlink-app
+      imageName: data61/anonlink-app
       jobName: 'anonlink_app'
       
 - stage: stage_benchmark_image_build
@@ -56,7 +56,7 @@ stages:
   - template: .azurePipeline/templateDockerBuildPush.yml
     parameters:
       folder: './benchmarking'
-      imageName: anonlink-benchmark
+      imageName: data61/anonlink-benchmark
       jobName: 'anonlink_benchmark'
       
 - stage: stage_docs_tutorial_image_build
@@ -66,7 +66,7 @@ stages:
   - template: .azurePipeline/templateDockerBuildPush.yml
     parameters:
       folder: './docs/tutorial'
-      imageName: anonlink-docs-tutorials
+      imageName: data61/anonlink-docs-tutorials
       jobName: 'anonlink_docs_tutorials'
 
 #- stage: stage_integration_tests
@@ -165,7 +165,7 @@ stages:
     # Prepare all the variables required during the scripts.
     - template: .azurePipeline/templateDockerTagFromBranch.yml
     - script: |
-        echo $(DOCKER_TAG)$(Build.SourceVersion)| head -c 40 | xargs -I@ echo "##vso[task.setvariable variable=DEPLOYMENT]@"
+        echo $(Build.SourceVersion)| head -c 40 | xargs -I@ echo "##vso[task.setvariable variable=DEPLOYMENT]@"
       displayName: 'Set some variables part 1'
     - script: |
         echo "##vso[task.setvariable variable=KUBECONFIGFILE]$(DownloadSecureFile.secureFilePath)"
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -272,7 +272,7 @@ stages:
     # Note that all the test resources (i.e. the pvc, the job and the pod has a label `deployment` set to $(DEPLOYMENT)
     - script: |
         echo "Delete all the test resources"
-        kubectl get pods,jobs,pvc -n=$(NAMESPACE) -l deployment=$(DEPLOYMENT)
+        kubectl delete pods,jobs,pvc -n=$(NAMESPACE) -l deployment=$(DEPLOYMENT)
       condition: always()
       env:
         KUBECONFIG: $(KUBECONFIGFILE)
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -41,12 +41,12 @@ stages:
     parameters:
       folder: '.'
       dockerFilePath: './frontend/Dockerfile'
-      imageName: $(frontendImageName)
+      imageName: anonlink-nginx
       jobName: 'anonlink_nginx'
   - template: .azurePipeline/templateDockerBuildPush.yml
     parameters:
       folder: './backend'
-      imageName: $(backendImageName)
+      imageName: anonlink-app
       jobName: 'anonlink_app'
       
 - stage: stage_benchmark_image_build
@@ -56,7 +56,7 @@ stages:
   - template: .azurePipeline/templateDockerBuildPush.yml
     parameters:
       folder: './benchmarking'
-      imageName: $(benchmarkingImageName)
+      imageName: anonlink-benchmark
       jobName: 'anonlink_benchmark'
       
 - stage: stage_docs_tutorial_image_build
@@ -66,7 +66,7 @@ stages:
   - template: .azurePipeline/templateDockerBuildPush.yml
     parameters:
       folder: './docs/tutorial'
-      imageName: $(tutorialImageName)
+      imageName: anonlink-docs-tutorials
       jobName: 'anonlink_docs_tutorials'
 
 #- stage: stage_integration_tests
@@ -84,6 +84,7 @@ stages:
 #    - template: .azurePipeline/templateDockerTagFromBranch.yml
 #    - script: |
 #        ./.azurePipeline/runDockerComposeTests.sh --no-ansi -p es$(DOCKER_TAG)$(Build.SourceVersion) -t $(DOCKER_TAG) -o $(resultFile) --type tests
+#      displayName: 'Start docker compose tests'
 #    - task: PublishTestResults@2
 #      condition: succeededOrFailed()
 #      inputs:
@@ -102,13 +103,14 @@ stages:
     timeoutInMinutes: 15
     variables:
       resultFile: results.json
-    displayName: Benchmarking
+    displayName: Benchmark
     pool:
       vmImage: 'ubuntu-16.04'
     steps:
     - template: .azurePipeline/templateDockerTagFromBranch.yml
     - script: |
         ./.azurePipeline/runDockerComposeTests.sh --no-ansi -p es$(DOCKER_TAG)$(Build.SourceVersion) -t $(DOCKER_TAG) -o $(resultFile) --type benchmark
+      displayName: 'Start docker compose benchmark'
     # Publish Pipeline Artifact
     # Publish a local directory or file as a named artifact for the current pipeline.
     - task: PublishPipelineArtifact@0
@@ -133,6 +135,7 @@ stages:
     - template: .azurePipeline/templateDockerTagFromBranch.yml
     - script: |
         ./.azurePipeline/runDockerComposeTests.sh --no-ansi -p es$(DOCKER_TAG)$(Build.SourceVersion) -t $(DOCKER_TAG) -o $(resultFile) --type tutorials
+      displayName: 'Start docker compose tutorials tests'
     - task: PublishTestResults@2
       condition: succeededOrFailed()
       inputs:
@@ -163,11 +166,13 @@ stages:
     - template: .azurePipeline/templateDockerTagFromBranch.yml
     - script: |
         echo "##vso[task.setvariable variable=DEPLOYMENT]deployment-$(DOCKER_TAG)$(Build.SourceVersion)"
+      displayName: 'Set some variables part 1'
     - script: |
         echo "##vso[task.setvariable variable=KUBECONFIGFILE]$(DownloadSecureFile.secureFilePath)"
         echo "##vso[task.setvariable variable=PVC]$(DEPLOYMENT)-test-results"
         echo $(backendImageName):$(DOCKER_TAG) | xargs -I@ echo "##vso[task.setvariable variable=IMAGE_NAME_WITH_TAG]@"
         echo $(DOCKER_TAG)-tmppod | xargs -I@ echo "##vso[task.setvariable variable=POD_NAME]@"
+      displayName: 'Set some variables part 2'
     # Start the whole service using helm. We are currently lucky, the helm version on azure is compatible with the one on our cluster, but the latest version of helm (Client 2.14.1) is not (tested from my laptop).
     - script: |
         echo "Start all the entity service pods"
@@ -186,6 +191,7 @@ stages:
                     --set workers.image.tag=$(DOCKER_TAG)
       env:
         KUBECONFIG: $(KUBECONFIGFILE)
+      displayName: 'Install the deployment on kubernetes'
     - script: |
         echo "Create Persistent Volume Claim"
         cat .azurePipeline/k8s_test_pvc.yaml.tmpl | \
@@ -194,10 +200,12 @@ stages:
         kubectl apply -n=$(NAMESPACE) -f -
       env:
         KUBECONFIG: $(KUBECONFIGFILE)
+      displayName: 'Create persistent volume claim for future results.'
     - script: |
         kubectl get services -n=$(NAMESPACE) -lapp=$(DEPLOYMENT)-entity-service -o jsonpath="{.items[0].spec.clusterIP}" | xargs -I@ echo "##vso[task.setvariable variable=SERVICE_IP]@"
       env:
         KUBECONFIG: $(KUBECONFIGFILE)
+      displayName: 'Set a variable for the service IP'
     - script: |
         echo "Start the test job"
         cat .azurePipeline/k8s_test_job.yaml.tmpl | \
@@ -210,6 +218,7 @@ stages:
         kubectl get pods -n=$(NAMESPACE) -l deployment=$(DEPLOYMENT) -o jsonpath="{.items[0].metadata.name}" | xargs -I@ kubectl logs -n=$(NAMESPACE) -f @
       env:
         KUBECONFIG: $(KUBECONFIGFILE)
+      displayName: 'Start the test job'
     - script: |
         echo "Get the test results"
         cat .azurePipeline/k8s_get_results.yaml.tmpl | \
@@ -222,6 +231,7 @@ stages:
         kubectl cp $(NAMESPACE)/$(POD_NAME):/mnt/results.xml k8s-results.xml
       env:
         KUBECONFIG: $(KUBECONFIGFILE)
+      displayName: 'Copy the results from the test job pod.'
     - task: PublishTestResults@2
       condition: succeeded()
       inputs:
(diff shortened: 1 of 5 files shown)


azure-pipelines.yml
```diff
@@ -160,55 +141,78 @@ stages:
...
+    - script: |
+        echo "Create Persistent Volume Claim"
+        cat .azurePipeline/k8s_test_pvc.yaml.tmpl | \
+        sed 's|\$PVC'"|$(PVC)|g" | \
+        sed 's|\$DEPLOYMENT_NAME'"|$(DEPLOYMENT)|g" | \
+        kubectl apply --wait=true -n=$(NAMESPACE) -f -
+      env:
+        KUBECONFIG: $(KUBECONFIGFILE)
+      displayName: 'Create persistent volume claim for future results.'
+    # Starting the whole entity-service
```
hardbyte: Idea for an additional step before deploying to a test cluster: lint/test the helm chart. E.g. https://helm.sh/blog/chart-testing-intro/#example-using-it-with-circleci

azure-pipelines.yml
```diff
@@ -160,55 +141,78 @@ stages:
...
+    # Task to get the IP of the service, which is require to run the test.
+    # Note that the port 80 is added at the end. This is in fact required for the command `dockerize -wait` to wait for a service before starting one.
     - script: |
-        kubectl get services -n=$(NAMESPACE) -lapp=$(DEPLOYMENT)-entity-service -o jsonpath="{.items[0].spec.clusterIP}" | xargs -I@ echo "##vso[task.setvariable variable=SERVICE_IP]@"
+        kubectl get services -n=$(NAMESPACE) -lapp=$(DEPLOYMENT)-entity-service -o jsonpath="{.items[0].spec.clusterIP}" | xargs -I@ echo "##vso[task.setvariable variable=SERVICE_IP]@:80"
       env:
         KUBECONFIG: $(KUBECONFIGFILE)
+      displayName: 'Set a variable for the service IP'
+    # The file `.azurePipeline/k8s_test_job.yaml.tmpl` is a template for the creation of the test job.
+    # The test job stops when all the tests have been ran, creating a test result file in the PVC.
```
hardbyte:
```suggestion
    # The test job stops when all the tests have been run, creating a test result file in the PVC.
```

azure-pipelines.yml
```diff
@@ -160,55 +141,78 @@ stages:
...
+        --set api.app.debug=true \
+        --set global.postgresql.postgresqlPassword=notaproductionpassword \
+        --set api.ingress.enabled=false \
+        --set api.www.image.tag=$(DOCKER_TAG) \
+        --set api.app.image.tag=$(DOCKER_TAG) \
+        --set api.dbinit.image.tag=$(DOCKER_TAG) \
+        --set workers.image.tag=$(DOCKER_TAG)
       env:
         KUBECONFIG: $(KUBECONFIGFILE)
+      displayName: 'Install the deployment on kubernetes'
```
hardbyte:
```suggestion
      displayName: 'Kubernetes deployment'
```

azure-pipelines.yml
```diff
@@ -160,55 +141,78 @@ stages:
...
+        kubectl apply --wait=true -n=$(NAMESPACE) -f -
+        kubectl get pods -n=$(NAMESPACE) -l deployment=$(DEPLOYMENT) -o jsonpath="{.items[0].metadata.name}" | xargs -I@ kubectl wait -n=$(NAMESPACE) --timeout=300s --for=condition=Ready pods/@
         kubectl get pods -n=$(NAMESPACE) -l deployment=$(DEPLOYMENT) -o jsonpath="{.items[0].metadata.name}" | xargs -I@ kubectl logs -n=$(NAMESPACE) -f @
       env:
         KUBECONFIG: $(KUBECONFIGFILE)
+      displayName: 'Start the test job'
+    # That is still something fun with kubernetes, we cannot copy a file from a stopped container in a pod (while this is possible with docker), which is
+    # why we are creating a persistent volume, running the test job using the volume, and finally starting a container to be up with the mounted
+    # volume and use `kubectl cp` on it while it still runs.
+    # This is tracked in https://github.com/kubernetes/kubectl/issues/454 but has been in April 2018, its priority raised in June 2019.
```
hardbyte: We could save the results to minio/s3 instead (which has to be configured or provisioned for the service to work).

azure-pipelines.yml
```diff
@@ -221,13 +225,16 @@ stages:
...
+      displayName: 'Copy the results from the test job pod.'
+    # Finally, we can publish the test results.
     - task: PublishTestResults@2
       condition: succeeded()
       inputs:
         testResultsFormat: 'JUnit'
         testResultsFiles: 'k8s-results.xml'
         testRunTitle: 'Publish kubernetes test results'
         failTaskOnFailedTests: true 
+    # And do a lot of cleaning.
```
hardbyte: All the cleaning could be done in one step if everything shared a label?

docs/devops.rst
```diff
@@ -0,0 +1,35 @@
+Devops
+===========
+
+Azure Pipeline
+--------------
+
+``entity-service`` is automatically built and tested using Azure Pipeline
```
hardbyte:
```suggestion
``anonlink-entity-service`` is automatically built and tested using Azure DevOps
```

docs/devops.rst
```diff
@@ -0,0 +1,35 @@
...
+in the project `Anonlink <https://dev.azure.com/data61/Anonlink>`.
+
+It consists only of a `build pipeline <https://dev.azure.com/data61/Anonlink/_build?definitionId=1>`.
+
+The build pipeline is described by the script `azurePipeline.yml`
```
hardbyte:
```suggestion
The build pipeline is defined by the script `azurePipeline.yml`
```

docs/devops.rst
```diff
@@ -0,0 +1,35 @@
...
+which is using resources from the folder `.azurePipeline`.
```
hardbyte:
```suggestion
which uses resources from the folder `.azurePipeline`.
```

docs/devops.rst
```diff
@@ -0,0 +1,35 @@
...
+Mainly, it:
+
+- builds and pushes docker images
+  - the frontend ``data61/anonlink-nginx``
+  - the backend ``data61/anonlink-app``
+  - the tutorials ``data61/anonlink-docs-tutorials`` (used to tests the tutorial Python Notebooks)
+  - the benchmark ``data61/anonlink-benchmark`` (used to run the benchmark)
+- runs the benchmark using ``docker-compose`` and publishes the artiafct in Azure
```
hardbyte:
```suggestion
- runs the benchmark using ``docker-compose`` and publishes the results as an artifact in Azure
```

docs/devops.rst
```diff
@@ -0,0 +1,35 @@
...
+
+The build pipeline is triggered for every push on every branch. It is not triggered by Pull
+Requests not to build twice the same commit.
+
+The build pipeline requires two environment variables provided by Azure environment:
+
+- `dockerHubId`: username for the pipeline to push images to Data61 dockerhub
+- `dockerHubPassword`: password for the corresponding username (this is a secret variable).
+
+It also requires a secret file (given in the Azure `Library` tab) containing the ``aws`` cluster
```
hardbyte:
```suggestion
It also requires a secret file (given in the Azure `Library` tab) containing the ``k8s`` cluster
```

Although we happen to be using AWS I don't think this is aws specific.

docs/devops.rst
```diff
@@ -0,0 +1,35 @@
+Devops
+===========
+
+Azure Pipeline
```
hardbyte:
```suggestion
Continuous Integration
```

azure-pipelines.yml
```diff
@@ -133,6 +111,7 @@ stages:
     - template: .azurePipeline/templateDockerTagFromBranch.yml
     - script: |
         ./.azurePipeline/runDockerComposeTests.sh --no-ansi -p es$(DOCKER_TAG)$(Build.SourceVersion) -t $(DOCKER_TAG) -o $(resultFile) --type tutorials
+      displayName: 'Start docker compose tutorials tests'
```
hardbyte:
```suggestion
      displayName: 'Test notebook tutorials'
```

azure-pipelines.yml
```diff
@@ -160,55 +141,78 @@ stages:
...
     - template: .azurePipeline/templateDockerTagFromBranch.yml
+    # We are selecting a deployment name to be unique, not to have collision when pushing multiple times on the same branch
+    # So we are using the git commit hash. BUT, a k8s deployment MUST start with a letter, so added `es` to start its name, 
+    # then it needs to be short enough not to fail other tags built from it. In fact, most of the created objects from the templates
+    # will use the deployment name as a prefix for their tags, adding extra suffixes. If the deployment name is too long, 
+    # the tags may be longer than 63 characters which can fail the deployment with some warnings hard to see.
     - script: |
-        echo "##vso[task.setvariable variable=DEPLOYMENT]deployment-$(DOCKER_TAG)$(Build.SourceVersion)"
+        echo es$(Build.SourceVersion)| head -c 10 | xargs -I@ echo "##vso[task.setvariable variable=DEPLOYMENT]@"
+      displayName: 'Set some variables part 1'
```
hardbyte:
```suggestion
      displayName: 'Set variables part 1'
```

azure-pipelines.yml
```diff
@@ -160,55 +141,78 @@ stages:
...
+    # The task setting variables is split in two, because we cannot re-use directly a set variable in the same task
+    # (in this case the value of $(DEPLOYMENT)).
     - script: |
         echo "##vso[task.setvariable variable=KUBECONFIGFILE]$(DownloadSecureFile.secureFilePath)"
         echo "##vso[task.setvariable variable=PVC]$(DEPLOYMENT)-test-results"
         echo $(backendImageName):$(DOCKER_TAG) | xargs -I@ echo "##vso[task.setvariable variable=IMAGE_NAME_WITH_TAG]@"
-        echo $(DOCKER_TAG)-tmppod | xargs -I@ echo "##vso[task.setvariable variable=POD_NAME]@"
-    # Start the whole service using helm. We are currently lucky, the helm version on azure is compatible with the one on our cluster, but the latest version of helm (Client 2.14.1) is not (tested from my laptop).
+        echo $(DEPLOYMENT)-tmppod | xargs -I@ echo "##vso[task.setvariable variable=POD_NAME]@"
+      displayName: 'Set some variables part 2'
```
hardbyte:
```suggestion
      displayName: 'Set variables part 2'
```

azure-pipelines.yml
```diff
@@ -160,55 +141,78 @@ stages:
...
+    # The file `.azurePipeline/k8s_test_pvc.yaml.tmpl` is a template for the creation of persistent volume via kubectl
+    - script: |
+        echo "Create Persistent Volume Claim"
+        cat .azurePipeline/k8s_test_pvc.yaml.tmpl | \
+        sed 's|\$PVC'"|$(PVC)|g" | \
+        sed 's|\$DEPLOYMENT_NAME'"|$(DEPLOYMENT)|g" | \
+        kubectl apply --wait=true -n=$(NAMESPACE) -f -
+      env:
+        KUBECONFIG: $(KUBECONFIGFILE)
+      displayName: 'Create persistent volume claim for future results.'
```
hardbyte:
```suggestion
      displayName: 'Create persistent volume claim for integration test results.'
```

azure-pipelines.yml
```diff
@@ -160,55 +141,78 @@ stages:
...
         sed 's|\$IMAGE_NAME_WITH_TAG'"|$(IMAGE_NAME_WITH_TAG)|g" | \
         sed 's|\$SERVICE_IP'"|$(SERVICE_IP)|g" | \
-        kubectl apply -n=$(NAMESPACE) -f -
-        kubectl get pods -n=$(NAMESPACE) -l deployment=$(DEPLOYMENT) -o jsonpath="{.items[0].metadata.name}" | xargs -I@ kubectl wait -n=$(NAMESPACE) --timeout=120s --for=condition=Ready pods/@
+        kubectl apply --wait=true -n=$(NAMESPACE) -f -
+        kubectl get pods -n=$(NAMESPACE) -l deployment=$(DEPLOYMENT) -o jsonpath="{.items[0].metadata.name}" | xargs -I@ kubectl wait -n=$(NAMESPACE) --timeout=300s --for=condition=Ready pods/@
         kubectl get pods -n=$(NAMESPACE) -l deployment=$(DEPLOYMENT) -o jsonpath="{.items[0].metadata.name}" | xargs -I@ kubectl logs -n=$(NAMESPACE) -f @
       env:
         KUBECONFIG: $(KUBECONFIGFILE)
+      displayName: 'Start the test job'
```
hardbyte:
```suggestion
      displayName: 'Run integration tests'
```

docs/devops.rst
```diff
@@ -0,0 +1,35 @@
...
+  - the frontend ``data61/anonlink-nginx``
+  - the backend ``data61/anonlink-app``
+  - the tutorials ``data61/anonlink-docs-tutorials`` (used to tests the tutorial Python Notebooks)
+  - the benchmark ``data61/anonlink-benchmark`` (used to run the benchmark)
+- runs the benchmark using ``docker-compose`` and publishes the artiafct in Azure
+- runs the tutorial tests using ``docker-compose`` and publishes the results in Azure
+- runs the integration tests by deploying the whole service on ``Kubernetes``, runs the tests and publishes the results in Azure.   
+
+The build pipeline is triggered for every push on every branch. It is not triggered by Pull
+Requests not to build twice the same commit.
```
hardbyte:
```suggestion
Requests to avoid duplicate testing and building potentially untrusted code.
```

azure-pipelines.yml
```diff
@@ -160,55 +141,78 @@ stages:
         secureFile: awsClusterConfig
     # Prepare all the variables required during the scripts.
     - template: .azurePipeline/templateDockerTagFromBranch.yml
+    # We are selecting a deployment name to be unique, not to have collision when pushing multiple times on the same branch
+    # So we are using the git commit hash. BUT, a k8s deployment MUST start with a letter, so added `es` to start its name, 
+    # then it needs to be short enough not to fail other tags built from it. In fact, most of the created objects from the templates
+    # will use the deployment name as a prefix for their tags, adding extra suffixes. If the deployment name is too long, 
+    # the tags may be longer than 63 characters which can fail the deployment with some warnings hard to see.
```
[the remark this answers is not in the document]
gusmith: That was fun to find...

azure-pipelines.yml
```diff
@@ -160,55 +141,78 @@ stages:
...
-        echo $(DOCKER_TAG)-tmppod | xargs -I@ echo "##vso[task.setvariable variable=POD_NAME]@"
-    # Start the whole service using helm. We are currently lucky, the helm version on azure is compatible with the one on our cluster, but the latest version of helm (Client 2.14.1) is not (tested from my laptop).
+        echo $(DEPLOYMENT)-tmppod | xargs -I@ echo "##vso[task.setvariable variable=POD_NAME]@"
+      displayName: 'Set some variables part 2'
+    # The file `.azurePipeline/k8s_test_pvc.yaml.tmpl` is a template for the creation of persistent volume via kubectl
+    - script: |
+        echo "Create Persistent Volume Claim"
+        cat .azurePipeline/k8s_test_pvc.yaml.tmpl | \
+        sed 's|\$PVC'"|$(PVC)|g" | \
+        sed 's|\$DEPLOYMENT_NAME'"|$(DEPLOYMENT)|g" | \
```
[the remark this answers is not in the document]
gusmith: Kubectl does not accept template sadly. So the I thought about few other solutions:
 - create a new helm template but seems too much to just start a job which is only internal (I don't think we are expecting others to run integration tests)
 - updating the current helm release to include this job as a post installation hook, but I do not want to change the helm template for the deployment as it should not include or mention this integration test job
 - create a bash script to create the good files (possibly using `sed` in it, but hide them from the pipeline), or even do the whole integration test job
 - use `sed` in the pipeline and create multiple steps

I found that using helm would be too much just for that.
Then, I didn't wanted to create bash script because:
 - I don't think it could be used anywhere else than here (so not need to generalize and add documentation of all the parameters, etc)
 - described this way still allow a dev to repeat each step easily by hand if required.

In conclusion, I went for simplicity :)

azure-pipelines.yml
gusmith: Do we want this in this PR?
I will add `helm lint` and `helm upgrade --dry-run` for the moment.
It seems the tool you are pointing to is doing a lot more than just checking that the template is valid (e.g. version updates). So we will need to define precisely what we would like to check.

azure-pipelines.yml
gusmith: :+1:

azure-pipelines.yml
gusmith: :+1:

azure-pipelines.yml
hardbyte: Adding what you've suggested is perfect for now. thanks

azure-pipelines.yml
gusmith: I opened issue #393 to track this. But let's have first something working and optimize later. Or do you think this is a priority?

azure-pipelines.yml
gusmith: I'll clean all the test resources (the PVC, job and pod) using a label they have in common, but will keep separate the deployment cleaning in a separate step.
They currently do not have a label in common.

docs/devops.rst
gusmith: :+1:

docs/devops.rst
gusmith: :+1:

docs/devops.rst
gusmith: :+1:

docs/devops.rst
gusmith: :+1:

docs/devops.rst
gusmith: That's right: this is the kubeconfig containing the credentials to our aws cluster.

docs/devops.rst
gusmith: I put Azure pipeline because I did not want to also write the documentation for the other CI we are using, but I was expecting that at one point we should. So I'll put Continuous integration as section, and `Azure DevOps` as sub-section.

azure-pipelines.yml
gusmith: :+1:

azure-pipelines.yml
gusmith: :+1:

azure-pipelines.yml
gusmith: :+1:

azure-pipelines.yml
gusmith: :+1:

azure-pipelines.yml
gusmith: :+1:

docs/devops.rst
gusmith: :+1:

gusmith:
\azr help

gusmith:
The build was failing because the cluster did not have enough resources to provide the required ones.
The k8s step failed when re-run because the variable $(DEPLOYMENT) became equal to `foo.v1`. I'm assuming that the `.v1` came from azure, but because of this, it failed as a deployment name should not include a `.`. So I added a commit to remove special characters in the deployment name.

PR #405, opened 2019-08-20
Feature connection pools

gusmith:
Simplify the access to dabases to a single way:
```
with DBConn() as conn:
  ...
```
It will under the hood handle how to get a connection: either using a connection pool if available, or creating a single use connection.

The two main entry points which behave differently are due to the application having different entry points: mainly one for celery and one for flask.
Each behaves quite differently (yeah...)
But with these changes a dev will not have to worry from where the method is called.

The PR may also look bigger than what it is: I removed some method to create databases connections, to use the single way, which was repeated on a number of files.

### DIFF
--- a/backend/entityservice/cache.py
+++ b/backend/entityservice/cache.py
@@ -7,7 +7,7 @@ import structlog
 
 from entityservice import serialization
 from entityservice import database
-from entityservice.database import logger
+from entityservice.database import logger, DBConn
 from entityservice.object_store import connect_to_object_store
 from entityservice.settings import Config as config
 
@@ -59,19 +59,20 @@ def get_deserialized_filter(dp_id):
         return pickle.loads(r.get(key))
     else:
         logger.debug("Looking up popcounts and filename from database")
-        db = database.connect_db()
-        mc = connect_to_object_store()
-        serialized_filters_file = database.get_filter_metadata(db, dp_id)
+        with DBConn() as db:
+            #db = database.connect_db()
+            mc = connect_to_object_store()
+            serialized_filters_file = database.get_filter_metadata(db, dp_id)
 
-        logger.debug("Getting filters from object store")
+            logger.debug("Getting filters from object store")
 
-        # Note this uses already calculated popcounts unlike
-        # serialization.deserialize_filters()
-        raw_data_response = mc.get_object(config.MINIO_BUCKET, serialized_filters_file)
-        python_filters = serialization.binary_unpack_filters(raw_data_response)
+            # Note this uses already calculated popcounts unlike
+            # serialization.deserialize_filters()
+            raw_data_response = mc.get_object(config.MINIO_BUCKET, serialized_filters_file)
+            python_filters = serialization.binary_unpack_filters(raw_data_response)
 
-        set_deserialized_filter(dp_id, python_filters)
-        return python_filters
+            set_deserialized_filter(dp_id, python_filters)
+            return python_filters
 
 
 def remove_from_cache(dp_id):
--- a/backend/entityservice/cache.py
+++ b/backend/entityservice/cache.py
@@ -60,19 +60,18 @@ def get_deserialized_filter(dp_id):
     else:
         logger.debug("Looking up popcounts and filename from database")
         with DBConn() as db:
-            #db = database.connect_db()
-            mc = connect_to_object_store()
             serialized_filters_file = database.get_filter_metadata(db, dp_id)
 
-            logger.debug("Getting filters from object store")
+        mc = connect_to_object_store()
+        logger.debug("Getting filters from object store")
 
-            # Note this uses already calculated popcounts unlike
-            # serialization.deserialize_filters()
-            raw_data_response = mc.get_object(config.MINIO_BUCKET, serialized_filters_file)
-            python_filters = serialization.binary_unpack_filters(raw_data_response)
+        # Note this uses already calculated popcounts unlike
+        # serialization.deserialize_filters()
+        raw_data_response = mc.get_object(config.MINIO_BUCKET, serialized_filters_file)
+        python_filters = serialization.binary_unpack_filters(raw_data_response)
 
-            set_deserialized_filter(dp_id, python_filters)
-            return python_filters
+        set_deserialized_filter(dp_id, python_filters)
+        return python_filters
 
 
 def remove_from_cache(dp_id):
--- a/backend/entityservice/database/util.py
+++ b/backend/entityservice/database/util.py
@@ -1,14 +1,16 @@
 import time
 
+import atexit
 import psycopg2
 import psycopg2.extras
-from flask import current_app, g
+from psycopg2.pool import ThreadedConnectionPool
+from flask import current_app
 from structlog import get_logger
 
-from entityservice import database
 from entityservice.settings import Config as config
 
 logger = get_logger()
+connection_pool = None
 
 
 def query_db(db, query, args=(), one=False):
@@ -25,29 +27,6 @@ def query_db(db, query, args=(), one=False):
     return (rv[0] if rv else None) if one else rv
 
 
-def connect_db():
-    db = config.DATABASE
-    host = config.DATABASE_SERVER
-    user = config.DATABASE_USER
-    pw = config.DATABASE_PASSWORD
-
-    logger.info("Trying to connect to database.", database=db, user=user, host=host)
-    # We nest the try/except blocks to allow one re-attempt with defaults
-    try:
-        try:
-            conn = psycopg2.connect(database=db, user=user, password=pw, host=host)
-        except psycopg2.OperationalError:
-            logger.warning("fallback connecting to default postgres db", database=db, user=user, host=host)
-            conn = psycopg2.connect(database='postgres', user=user, password=pw, host=host)
-
-    except psycopg2.Error:
-        logger.warning("Can't connect to database")
-        raise ConnectionError("Issue connecting to database")
-
-    logger.debug("Database connection established")
-    return conn
-
-
 def init_db(delay=0.5):
     """
     Initializes the database by creating the required tables.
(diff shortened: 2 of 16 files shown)


hardbyte:
I'll try review this tomorrow or at worst Friday. I'll leave a link to a comment for myself to pick up - https://github.com/data61/anonlink-entity-service/issues/385#issuecomment-501055807

gusmith:
First PR I'm doing for Wilko to review in this case :)

backend/entityservice/settings.py
```diff
@@ -33,6 +33,7 @@ class Config(object):
     DATABASE = os.getenv('DATABASE', 'postgres')
     DATABASE_USER = os.getenv('DATABASE_USER', 'postgres')
     DATABASE_PASSWORD = os.getenv('DATABASE_PASSWORD', '')
+    DATABASE_MAX_CONNECTIONS = os.getenv('DATABASE_MAX_CONNECTIONS', '5')
```
wilko77: is that a good default value? How many concurrent connections do we typically have in a worker?

backend/entityservice/tasks/stats.py
```diff
@@ -1,37 +1,38 @@
...
 from entityservice.database import get_total_comparisons_for_project
 from entityservice.database import insert_comparison_rate
 
 
 @celery.task(ignore_result=True)
 def calculate_comparison_rate():
-    dbinstance = connect_db()
-    logger.info("Calculating global comparison rate")
+    with DBConn() as dbinstance:
+        #dbinstance = connect_db()
```
wilko77: remove

backend/entityservice/database/util.py
```diff
@@ -25,52 +27,103 @@ def query_db(db, query, args=(), one=False):
...
+                except psycopg2.Error:
+                    logger.warning("Can't connect to database")
+                    raise ConnectionError("Issue connecting to database")
+            else:
+                logger.debug("Get a database connection without a pool.")
+                self.__connect_db()
+                self.return_to_pool = False
+
+        else:
+            # Not sure where the connection is coming from...
```
wilko77: this is an artifact that is not needed anymore. delete the optional `conn` parameter and this else clause.

backend/entityservice/database/util.py
```diff
@@ -25,52 +27,103 @@ def query_db(db, query, args=(), one=False):
...
+        if conn is None:
+            if connection_pool is not None:
+                logger.debug("Get a database connection from the pool.")
+                try:
+                    self.conn = connection_pool.getconn()
+                except psycopg2.Error:
+                    logger.warning("Can't connect to database")
+                    raise ConnectionError("Issue connecting to database")
+            else:
+                logger.debug("Get a database connection without a pool.")
```
wilko77: why do we need this fallback? I prefer that we force the use of connection pools. Otherwise it is hard to control the max number of connections.

gusmith:
Thanks @wilko77 for the really fast and deep review.
I'll answer and work on your comments as soon as I finish some observations which I'm adding to #385

gusmith:
@wilko77 , in the issue #385, I tested it using a real deployment using `docker-compose` on my local machine.
I show the number of active connection from the branch master, which increases after the benchmarks timeout, more specifically, after the benchmark deletes the run, creating `DBResourceMissing` errors which may have been badly handled and not closing their connections.
Doing exactly the same from this branch, the number of observed active connections stayed **exactly** the same at a really low number (1s frequency of observation).

gusmith:
From some observations, we have done a too good job: the workers are opening only one connection each and returning it. 
so having a max of 2 for the worker should be enough and a min of 1.
But I don't know if and when we could have more than 1 in the backend. 
I think the default of min = 1 and max = 1 should be good start. We should not forget that it is per fork worker. For the backend, I may configure it to have a max of 5. I'm mainly wondering how many connections are required on the deployed service on k8s.

backend/entityservice/database/util.py
```diff
@@ -25,52 +27,103 @@ def query_db(db, query, args=(), one=False):
...
-    except psycopg2.Error:
-        logger.warning("Can't connect to database")
-        raise ConnectionError("Issue connecting to database")
-
-    logger.debug("Database connection established")
-    return conn
+            connection_pool = ThreadedConnectionPool(0, db_max_connections, database=db, user=user, password=pw, host=host)
+        except psycopg2.Error:
+            logger.warning("Can't connect to database")
+            raise ConnectionError("Issue connecting to database")
```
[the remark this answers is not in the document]
gusmith: :+1:

backend/entityservice/database/util.py
```diff
@@ -81,7 +134,20 @@ def __exit__(self, exc_type, exc_val, traceback):
                 # It was a Postgres exception
                 logger.warning(f"{exc_val.diag.severity} - {exc_val.pgerror}")
             # Note if we return True we swallow the exception, False we propagate it
-            return False
+            result = False
+            self.conn.cancel()
+
+        if self.return_to_pool:
+            if self.conn.closed != 0:
```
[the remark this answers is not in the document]
gusmith: :+1:

backend/entityservice/database/util.py
```diff
@@ -25,52 +27,103 @@ def query_db(db, query, args=(), one=False):
...
-            logger.warning("fallback connecting to default postgres db", database=db, user=user, host=host)
-            conn = psycopg2.connect(database='postgres', user=user, password=pw, host=host)
-
-    except psycopg2.Error:
-        logger.warning("Can't connect to database")
-        raise ConnectionError("Issue connecting to database")
-
-    logger.debug("Database connection established")
-    return conn
+            connection_pool = ThreadedConnectionPool(0, db_max_connections, database=db, user=user, password=pw, host=host)
```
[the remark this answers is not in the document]
gusmith: Added, default min to 1, and default max to 3. I observed that the celery worker only needs 1 connection, but flask sometimes would like to have more.

backend/entityservice/settings.py
gusmith: Workers need only 1. But flask sometimes need more. I will need to check what is returned by the server when it happens. A 500 would be good, asking the client to retry.

backend/entityservice/tasks/stats.py
gusmith: :+1:

backend/entityservice/database/util.py
gusmith: I am only aware of two entry points to the application. I do not know if there were some other use cases which would require a single connection. I'll update it forcing the use of the pools, and raising an exception to init the pool if not done.

backend/entityservice/database/util.py
gusmith: :+1:

backend/entityservice/settings.py
gusmith: Tested, it return well a 500. Returning a 503 would be better, but that's a good start.

gusmith:
This should close the issue #385 when merged in.
Flask may still raise an exception if the connection pool is empty, returning a 500 to the client, who could just retry later as the pool should fill back.

backend/entityservice/cache.py
```diff
@@ -59,19 +59,20 @@ def get_deserialized_filter(dp_id):
...
-        python_filters = serialization.binary_unpack_filters(raw_data_response)
+            # Note this uses already calculated popcounts unlike
+            # serialization.deserialize_filters()
+            raw_data_response = mc.get_object(config.MINIO_BUCKET, serialized_filters_file)
+            python_filters = serialization.binary_unpack_filters(raw_data_response)
 
-        set_deserialized_filter(dp_id, python_filters)
-        return python_filters
+            set_deserialized_filter(dp_id, python_filters)
+            return python_filters
```
wilko77: only the `serialized_filters_file = database.get_filter_metadata(db, dp_id)` line actually needs the database connection. Its better to return it to the pool straight away and not wait for all this minio stuff to finish first.

backend/entityservice/cache.py
```diff
@@ -59,19 +59,20 @@ def get_deserialized_filter(dp_id):
         return pickle.loads(r.get(key))
     else:
         logger.debug("Looking up popcounts and filename from database")
-        db = database.connect_db()
-        mc = connect_to_object_store()
-        serialized_filters_file = database.get_filter_metadata(db, dp_id)
+        with DBConn() as db:
+            #db = database.connect_db()
```
wilko77: remove commented code.

backend/entityservice/tasks/stats.py
```diff
@@ -1,37 +1,37 @@
...
 
-        with dbinstance.cursor() as cur:
-            insert_comparison_rate(cur, rate)
+            with dbinstance.cursor() as cur:
+                insert_comparison_rate(cur, rate)
 
-        dbinstance.commit()
-    else:
-        logger.warning("Can't compute comparison rate yet")
+            dbinstance.commit()
```
wilko77: that's now unnecessary. When `DBConn()` exits, it commits.

backend/entityservice/tasks/project_cleanup.py
```diff
@@ -17,14 +18,15 @@ def remove_project(project_id):
...
-    conn = db.connect_db()
-    log.debug("Deleting project resourced from database")
-    db.delete_project_data(conn, project_id)
-    log.debug("Getting object store files associated with project from database")
-    object_store_files = db.get_all_objects_for_project(conn, project_id)
-    log.debug(f"Removing {len(object_store_files)} object store files associated with project.")
-    delete_minio_objects.delay(object_store_files, project_id)
-    log.info("Project resources removed")
+    with DBConn() as conn:
+        #conn = db.connect_db()
```
wilko77: we don't need that line any more. It doesn't spark joy. Thank it for all the services it has done for us and then send it of to new adventures.

backend/entityservice/cache.py
gusmith: :+1:

backend/entityservice/cache.py
gusmith: :+

backend/entityservice/tasks/stats.py
gusmith: Nicely spotted.

PR #406, opened 2019-08-20
Benchmark repeat parameter and results pushed to s3

gusmith:
Two main features added at the same time (they could have been split).
Use of boto3 for the benchmark to be able to push the results to s3 instead of relying on k8s volumes which are hard to access afterward.
A parameter `repetition` has been added to an experiment schema to be able to repeat the same experiment a number of time in the same benchmarking k8s job. This should reduce the number of issue we could observe when running 100 times 1M*1M linkage as we were previously repeating the same k8s jobs 100 times, which can have issues with volumes which cannot be shared.

### DIFF
--- a/benchmarking/benchmark.py
+++ b/benchmarking/benchmark.py
@@ -272,6 +272,7 @@ def run_single_experiment(server, config, threshold, size_a, size_b, experiment)
     result = {}
     credentials = {}
     run = {}
+    logger.info("Starting time: {}".format(time.asctime()))
     try:
         credentials = rest_client.project_create(server, config['schema'], 'mapping',
                                                  "benchy_{}".format(experiment))
@@ -305,6 +306,7 @@ def run_single_experiment(server, config, threshold, size_a, size_b, experiment)
         logger.info('cleaning up...')
         delete_resources(config, credentials, run)
 
+    logger.info("Ending time: {}".format(time.asctime()))
     return result
 
 
--- a/benchmarking/benchmark.py
+++ b/benchmarking/benchmark.py
@@ -66,7 +66,7 @@ def read_config():
     DEFAULT_DATA_FOLDER = './data'
     DEFAULT_RESULTS_PATH = 'results.json'
 
-    DEFAULT_S3_BUCKET = 'anonlink-benchmark-result'
+    DEFAULT_OBJECT_STORE_BUCKET = 'anonlink-benchmark-result'
 
     try:
         server = os.getenv('SERVER')
@@ -89,18 +89,20 @@ def read_config():
             'results_path': results_path
         }
 
-        if 'AWS_ACCESS_KEY_ID' in os.environ:
-            aws_access_key_id = os.getenv('AWS_ACCESS_KEY_ID')
-            aws_secret_access_key = os.getenv('AWS_SECRET_ACCESS_KEY')
-            s3_bucket = os.getenv('S3_BUCKET', DEFAULT_S3_BUCKET)
+        if 'OBJECT_STORE_ACCESS_KEY' in os.environ:
+            object_store_server = os.getenv('OBJECT_STORE_SERVER')
+            object_store_access_key = os.getenv('OBJECT_STORE_ACCESS_KEY')
+            object_store_secret_key = os.getenv('OBJECT_STORE_SECRET_KEY')
+            object_store_bucket = os.getenv('OBJECT_STORE_BUCKET', DEFAULT_OBJECT_STORE_BUCKET)
             config.update({
-                'push_to_s3': True,
-                'aws_access_key_id': aws_access_key_id,
-                'aws_secret_access_key': aws_secret_access_key,
-                's3_bucket': s3_bucket
+                'push_to_object_store': True,
+                'object_store_server': object_store_server,
+                'object_store_access_key': object_store_access_key,
+                'object_store_secret_key': object_store_secret_key,
+                'object_store_bucket': object_store_bucket
             })
         else:
-            config['push_to_s3'] = False
+            config['push_to_object_store'] = False
 
         return config
     except Exception as e:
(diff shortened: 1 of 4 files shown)


benchmarking/benchmark.py
```diff
@@ -230,46 +235,71 @@ def run_experiments(config):
...
+                                                 "benchy_{}".format(experiment))
+        # upload clks
+        upload_binary_clks(config, size_a, size_b, credentials)
+        # create run
+        run = rest_client.run_create(server, credentials['project_id'], credentials['result_token'],
+                                     threshold,
+                                     "{}_{}".format(experiment, threshold))
+        # wait for result
+        run_id = run['run_id']
+        result['run_id'] = run_id
```
hardbyte: Nice if you store `project_id` as well this PR would also close https://github.com/data61/anonlink-entity-service/issues/326

benchmarking/benchmark.py
```diff
@@ -230,46 +235,71 @@ def run_experiments(config):
...
+        logger.info('cleaning up...')
+        delete_resources(config, credentials, run)
+
+    return result
+
+
+def push_result_s3(experiment_file):
+    client = boto3.client(
+        's3',
+        aws_access_key_id=os.getenv('AWS_ACCESS_KEY_ID'),
```
hardbyte: My hunch is we should be agnostic as multiple services offer the S3 api (including MINIO which we already use in this project). 

How about `OBJECT_STORE_SERVER`, `OBJECT_STORE_ACCESS_KEY`,  `OBJECT_STORE_BUCKET` etc as the environment variables? And open an issue to change them in the backend settings from `MINIO_` - https://github.com/data61/anonlink-entity-service/blob/develop/backend/entityservice/settings.py#L27

benchmarking/benchmark.py
```diff
@@ -230,46 +235,71 @@ def run_experiments(config):
...
+        aws_secret_access_key=os.getenv('AWS_SECRET_ACCESS_KEY')
+    )
+    s3_bucket = "anonlink-benchmark-result"
+    client.upload_file(experiment_file, s3_bucket, "results.json")
```
hardbyte: So if this benchmarking job *is* run multiple times in k8s with the same s3 bucket (extremely likely) the results get overridden. I suggest we include a timestamp or uuid in the uploaded filename.

benchmarking/benchmark.py
```diff
@@ -293,6 +323,7 @@ def main():
         pprint(results)
         with open(config['results_path'], 'wt') as f:
             json.dump(results, f)
+        push_result_s3(config['results_path'])
```
hardbyte: Should uploading to s3 be optional? Someone might want to run this benchmarking locally and just see the output as before?

benchmarking/benchmark.py
gusmith: :+1:

benchmarking/benchmark.py
gusmith: I'm actually not convinced. My reasoning being the following:
 - we do not support anything else than aws in the benchmark
 - the benchmark is fully independent from the main repository, so variable names from the main repo should not impact choices made for the benchmark (even if consistency may be a good idea)
 - but if we are thinking about both repos together, I would actually made them even more specific, such as `AWS_BENCHMARK_...` to be sure that we cannot imagine they will be used for any other context than the chosen one. I would also point out that if we implement a script doing all at once (deploying and benchmarking), it's good to use different environment variable names to be sure that a mis-configuration will not lead to misuse of a token. E.g. I create a script deploying the entity service setting the env var `OBJECT_STORE_ACCESS_KEY`, if the script also starts the benchmark but I forgot to update the env var, I would push the results to the same bucket. Here it is not important as we are not deleting anything but if we were, we could do really bad things...
 - would there be a scenario where a single service has multiple keys? E.g. one key for the bucket `x` and one key for the bucket `y`? The underlying question being if we prefer to have keys per application, or keys per use-case (the application can do x, y and z, or key a can do x, key b can do y and key c can do z, and I give the application the keys a, b and c). Both have pros and cons, but we may want to think about it
- if we generalise to `OBJECT_STORE_SERVER`, I think we will also need an extra field `OBJECT_STORE_TYPE`, in which case we may also need to modify the description of the env variables: one object store type may not use a bucket and access key, but something totally different (I have no clue what it could be, but maybe :) ). This field would also help to know which service to use. 

But the more options we are adding, the more I would assume we should push them into a command line tool instead of env vars.

While not being convinced, I'm also not really strongly opinionated on this, so happy to change if that's the preference :)

benchmarking/benchmark.py
gusmith: :+1: 
I will use timestamps, to more easily access them after (the job may not be kept on k8s, so the UUID may become hard to access).

benchmarking/benchmark.py
gusmith: :+1: 
I'll just check if the environment variables have been set, not adding an extra one.

gusmith:
@hardbyte in the benchmark, we are printing out the environment variables if the is an exception thrown in the `read_conf` method. Do you know where the benchmark logs are pushed to? Mainly: who can see them? If we add some credentials, we may not want to print them...

gusmith:
@hardbyte all your concerns have been resolved, except the naming of the environment variables, for which we may want to chat, cf discussion.

benchmarking/benchmark.py
hardbyte: I'm not too fussed with the variable names but it would be nice to include a way to set the object store server so the benchmark user can decide for themselves.

See the docs at https://docs.min.io/docs/minio-select-api-quickstart-guide.html it looks like all you would need is to add an `endpoint_url` which is only set if an environment variable is present e.g. `S3_SERVER` or my preference `OBJECT_STORE_SERVER`.

```python
s3 = boto3.client('s3',
                  endpoint_url='http://localhost:9000',
                  aws_access_key_id='minio',
                  aws_secret_access_key='minio123',
                  )
```

I don't think we need to tell the benchmark component the `TYPE` of object store - that would defeat the purpose of an abstraction - in this case the S3 API.

gusmith: `OBJECT_STORE_SERVER` added, and other ones renamed to: `OBJECT_STORE_ACCESS_KEY`, `OBJECT_STORE_SECRET_KEY` and `OBJECT_STORE_BUCKET`

Issue #408, opened 2019-08-23
Database disconnection is not recovered well by deployment/benchmark


gusmith:
This issue is opened from observations when running 100 times 1M*1M linkage on a deployed service on k8s.
At one point, the database fails and restarts (see issue #407 ).

From his point, the whole deployment melt-down: the workers restarts a number of times (I cannot see the source log with their failure), the flask pod restarts.
Flask seems to be OK when the database is not available (raising a lot of exception, but that sounds expected). But at the end and without clear notice:
```
[2019-08-22 18:54:22 +0000] [1] [INFO] Handling signal: term
[2019-08-22 18:54:22 +0000] [10] [INFO] Worker exiting (pid: 10)
[2019-08-22 18:54:22 +0000] [11] [INFO] Worker exiting (pid: 11)
[2019-08-22 18:54:22 +0000] [12] [INFO] Worker exiting (pid: 12)
```

hardbyte:
Might be useful to look at automated retrying with an exponential backoff before failure.

PR #409, opened 2019-08-26
Fix exception log

gusmith:
To print an exception in a log, the line 
```
logger.warning("Can't connect to database", e)
```
fails raising: "TypeError: not all arguments converted during string formatting"
Instead, as we are mainly printing the exception in a `try except` context, use the param `exc_info=True` to print it.

### DIFF
--- a/backend/entityservice/database/util.py
+++ b/backend/entityservice/database/util.py
@@ -103,8 +103,8 @@ class DBConn:
         return self.conn
 
     def __exit__(self, exc_type, exc_val, traceback):
+        result = True
         try:
-            result = True
             if not exc_type:
                 self.conn.commit()
                 for notice in self.conn.notices:
@@ -117,11 +117,11 @@ class DBConn:
                 # Note if we return True we swallow the exception, False we propagate it
                 result = False
                 self.conn.cancel()
-            return result
         except Exception as e:
             logger.warning("Exception cleaning a connection before closing it or returning it to the pool.", e)
         finally:
             connection_pool.putconn(self.conn, close=False)
+            return result
 
 
 def execute_returning_id(cur, query, args):
--- a/backend/entityservice/database/util.py
+++ b/backend/entityservice/database/util.py
@@ -54,11 +54,11 @@ def init_db_pool():
 
     global connection_pool
     if connection_pool is None:
-        logger.info("Initialize the database connection pool.", database=db, user=user, password=pw, host=host)
+        logger.info("Initialize the database connection pool.")
         try:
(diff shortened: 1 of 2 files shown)


backend/entityservice/database/util.py
```diff
@@ -54,11 +54,11 @@ def init_db_pool():
 
     global connection_pool
     if connection_pool is None:
-        logger.info("Initialize the database connection pool.", database=db, user=user, password=pw, host=host)
+        logger.info("Initialize the database connection pool.")
```
wilko77: why do you delete the extra info? I agree that revealing the password might not be such a great idea, but the rest might actually be helpful in some cases...

backend/entityservice/database/util.py
```diff
@@ -54,11 +54,11 @@ def init_db_pool():
...
         try:
             connection_pool = ThreadedConnectionPool(db_min_connections, db_max_connections, database=db, user=user, password=pw, host=host)
         except psycopg2.Error as e:
-            logger.warning("Can't connect to database")
+            logger.warning("Can't connect to database", exc_info=True)
```
wilko77: There is also https://docs.python.org/3.7/library/logging.html#logging.Logger.exception which is a neat wrapper around `logger.error(msg, exc_info=True)`. Might look cleaner.
I disagree with the log level. This should at least be ERROR, as it affects the correct workings of the system in a significant way. Well, nothing's gonna work properly without a db connection... 
Same with most of the other log level in this PR.

gusmith: I indeed saw this method, but wrapping around the log level `error`. We were currently using `warning`, and my thought is that it should only be a warning for the entity service app (at this level): depending on the context, we may want to do something different. It is not yet an error as we are raising another exception which should become and error if not handled. 
For example, if I request the entity service status which call the db to get the number of runs, if the database is not reachable, the application just need to return a 503 to the caller, asking to send the request back in a short time. The app is not in a wrong state. But when a task is trying to write something to the db and the db is not reachable, that may be logged as an error.

I would also argue that this specific service should handle the lost of connection to any other service it is using as it can, and will happen (e.g. the other service restarted, the database is reaching its bandwidth limits...). We are currently relying too much on the fact that everything will be up when requested. 

In conclusion, I would even argue that it should only be a debug message here or even simply removed, but everytime we are using the DBConn context, we should explicitely say what to do if a `ConnectionError` occurs. This is slightly mentioned in the issue #408 

But in the meanwhile and because we are not handling any of these, I'm agreeing with you that it should be an error.

backend/entityservice/database/util.py
gusmith: :+1:

PR #413, opened 2019-09-03
Feature delete runs

hardbyte:
Removes the warning about run deletion not being implemented.

Note it actually deletes the row in the `runs` table, and other tables' rows get deleted via the cascade mechanism. I ran this many times during a run to see how the backend deals with the database changing under it and the tasks correctly log the Database Integrity Error when they attempt to update a now missing row.


backend/entityservice/database/deletion.py
```diff
@@ -3,6 +3,27 @@
...
+            # and permutation_masks should delete their rows when
+            # runs row is deleted due to "on DELETE CASCADE"
+            cur.execute("""
+                DELETE
+                FROM runs
+                WHERE
+                    run_id = %s
+                """, [run_id])
+        logger.debug("Finishing DB transaction to delete run")
+    logger.info("Run resources removed")
```
gusmith: Should it include the runID?

backend/entityservice/database/selections.py
```diff
@@ -387,7 +394,19 @@ def get_all_objects_for_project(db, project_id):
...
+            SELECT 
+              similarity_scores.file
+            FROM 
+              similarity_scores
+            WHERE 
+              similarity_scores.run = %s
+            """, [run_id])
+    similarity_files = [res['file'] for res in query_response]
+    if len(similarity_files):
+        return similarity_files[0]
```
gusmith: Should we log something if there is more than 1 file? I know it should not happen, but what if?

backend/entityservice/tests/test_deletes.py
```diff
@@ -0,0 +1,45 @@
+from entityservice.tests.util import post_run, delete_project, create_project_upload_fake_data, delete_run, \
+    get_project_description, get_run, wait_for_run_completion
+import pytest
+
+
+def test_delete_project_after_creating_run_with_clks(requests, result_type_number_parties):
+    result_type, number_parties = result_type_number_parties
+    project, _ = create_project_upload_fake_data(
+        requests, [100] * number_parties, overlap=0.5, result_type=result_type)
+    post_run(requests, project, 1.0)
```
gusmith: The threshold was 0.9, why the change to 1? Is it to create the result faster?

backend/entityservice/tests/test_deletes.py
```diff
@@ -0,0 +1,45 @@
+from entityservice.tests.util import post_run, delete_project, create_project_upload_fake_data, delete_run, \
+    get_project_description, get_run, wait_for_run_completion
+import pytest
+
+
+def test_delete_project_after_creating_run_with_clks(requests, result_type_number_parties):
+    result_type, number_parties = result_type_number_parties
+    project, _ = create_project_upload_fake_data(
+        requests, [100] * number_parties, overlap=0.5, result_type=result_type)
```
gusmith: Instead of the three first lies, could use `_create_data_linkage_run(requests, result_type_number_parties)`

backend/entityservice/tests/test_deletes.py
```diff
@@ -0,0 +1,45 @@
...
+    _checks_after_run_deleted(requests, project, run_id)
+
+
+def test_delete_runs_after_completed_run(requests, result_type_number_parties):
+    project, run_id = _create_data_linkage_run(requests, result_type_number_parties)
+    result_token = project['result_token']
+    wait_for_run_completion(requests, project, run_id, result_token, timeout=20)
+    delete_run(requests, project['project_id'], run_id, result_token)
+    _checks_after_run_deleted(requests, project, run_id)
+    delete_run(requests, project['project_id'], run_id, result_token)
```
gusmith: Why is it done twice?

backend/entityservice/tests/util.py
```diff
@@ -196,17 +196,27 @@ def create_project_upload_data(
     return new_project_data, json_responses
 
 
+def _check_delete_response(r):
+    # Note we allow for a 403 because the project may already have been deleted
```
gusmith: Just wondering: why not 404? Not part at all with this PR as you haven't changed anything about it, but I'm seeing in the swagger file that we can return a 403 or a 404. Now, if a run has been deleted, I would assume this is not available, so we should return a 404, but as we are also requesting access token, and a run does not have any if not present, should return indeed a 403. So the question would be: is there any cases where we could return a 404 on the `/runs` endpoints?

backend/entityservice/views/run/description.py
```diff
@@ -2,28 +2,46 @@
...
 
 
 def get(project_id, run_id):
     log = logger.bind(pid=project_id, rid=run_id)
-    logger.info("request description of a run")
-    # Check the project and run resources exist
-    abort_if_run_doesnt_exist(project_id, run_id)
+    log.info("request description of a run")
+    authorize_run_detail(project_id, run_id)
+    log.debug("request for run description authorized")
```
gusmith: Should this log  be in the method `authorize_run_detail` instead of here?

backend/entityservice/views/run/description.py
```diff
@@ -2,28 +2,46 @@
...
 
     return RunDescription().dump(run_object)
 
 
 def delete(project_id, run_id):
     log = logger.bind(pid=project_id, rid=run_id)
-    log.warn("delete on runs is not implemented! Help!")
+    log.info("request to delete run")
+    authorize_run_detail(project_id, run_id)
+    log.info("approved request to delete run")
```
gusmith: Should be a debug message.

backend/entityservice/views/run/description.py
```diff
@@ -2,28 +2,46 @@
...
+
+    with db.DBConn() as conn:
+        log.debug("Retrieving run details from database")
+        similarity_file = get_similarity_file_for_run(conn, run_id)
+        if similarity_file:
+            log.debug(f"Queuing task to remove similarities file from object store")
```
gusmith: I don't think this string need to be formatted.

backend/entityservice/database/deletion.py
hardbyte: Yes it should - I'll update

backend/entityservice/tests/test_deletes.py
hardbyte: Yeah just to have the tests run faster

backend/entityservice/tests/test_deletes.py
hardbyte: To check that deleting an already deleted resource doesn't cause a 🔥

backend/entityservice/tests/util.py
hardbyte: This was to prevent enumeration on the `/runs` endpoint - overkill with our large UUID addresses. I think we only use 404 for wrong address e.g. `/links` instead of `/runs`

backend/entityservice/views/run/description.py
hardbyte: It could - I kept it separate because the `authorize_run_detail` can authorize either deletion or getting the run description.

PR #414, opened 2019-09-03
Use CHUNK_SIZE_AIM instead of old environment variables in compose

hardbyte:


tools/docker-compose.yml
```diff
@@ -68,11 +68,7 @@ services:
       - CELERY_ACKS_LATE=true
       - REDIS_USE_SENTINEL=false
       - CELERYD_MAX_TASKS_PER_CHILD=4096
-      - SMALL_COMPARISON_CHUNK_SIZE=100000000
-      - LARGE_COMPARISON_CHUNK_SIZE=1000000000
-      - SMALL_JOB_SIZE=100000000
-      - LARGE_JOB_SIZE=100000000000
-      - MATCH_THRESHOLD=0.98
```
gusmith: Why did you remove the default threshold?

hardbyte: Same reason - it isn't a used environment variable so it is just misleading people who might change it expecting something about our software to change...

backend/entityservice/settings.py
```diff
@@ -64,7 +64,7 @@ class Config(object):
     CELERY_ACKS_LATE = os.getenv('CELERY_ACKS_LATE', 'false') == 'true'
 
     # Number of comparisons per chunk (on average).
-    CHUNK_SIZE_AIM = int(os.getenv('CHUNK_SIZE_AIM', '300000000'))
+    CHUNK_SIZE_AIM = int(os.getenv('CHUNK_SIZE_AIM', '300_000_000'))
```
gusmith: I've haven't read the PEP 515 before :) Thanks for the learning.

Issue #419, opened 2019-09-05
Delete run leading to server error


hardbyte:
This race condition occured while running the integration tests [here](https://dev.azure.com/data61/Anonlink/_build/results?buildId=335&view=ms.vss-test-web.build-test-results-tab).

At first glance it seems like the celery worker correctly threw a DB exception in the `compute_filter_similarity` task, leading to the callback throwing an exception... so far this is fine. I note that the celery worker then tries to write to the runs table again to mark the run as failed:

```SQL
UPDATE runs SET
              state = 'error',
              time_completed = now()
WHERE ...
```

Over in the flask app we queue a task to delete any similarity files **then** start a db transaction to delete the run. It is conceivable that the delete files task finishes **before** a file is written.

Lastly in the flask app, we don't catch the DB integrity error and end up giving the user a 500 response :-/

Perhaps this could be addressed more like project deletion - by setting the run as `deleted` in the database instead of deleting the row. 

Note I didn't get to see the high memory worker logs - see #418 

## Logs

Integration test failure:

```
 ============================= test session starts ==============================
platform linux -- Python 3.7.3, pytest-3.5.1, py-1.8.0, pluggy-0.6.0
rootdir: /var/www, inifile:
plugins: celery-4.3.0
collected 371 items
entityservice/tests/test_admin.py ..                                     [  0%]
entityservice/tests/test_deletes.py ..............F
=================================== FAILURES ===================================
______ test_delete_runs_after_completed_run[result_type_number_parties2] _______
requests = &lt;requests.sessions.Session object at 0x7ff4d5ebdfd0&gt;
result_type_number_parties = ('permutations', 2)
    def test_delete_runs_after_completed_run(requests, result_type_number_parties):
        project, run_id = _create_data_linkage_run(requests, result_type_number_parties)
        result_token = project['result_token']
        wait_for_run_completion(requests, project, run_id, result_token, timeout=20)
&gt;       delete_run(requests, project['project_id'], run_id, result_token)
entityservice/tests/test_deletes.py:23: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
entityservice/tests/util.py:219: in delete_run
    _check_delete_response(r)
[...]

PR #420, opened 2019-09-06
Retry delete run on db failure

hardbyte:
This won't definitively close the root cause of issue #419 but should at least handle the flaky behavior.


backend/entityservice/views/run/description.py
```diff
@@ -1,5 +1,7 @@
+import psycopg2
 from flask import request
 from structlog import get_logger
+from retrying import retry
```
wilko77: retrying is unmaintained for more than a year now. https://github.com/rholder/retrying/issues/79
better: use tenacity https://github.com/jd/tenacity

backend/entityservice/views/run/description.py
```diff
@@ -23,21 +25,33 @@ def get(project_id, run_id):
     return RunDescription().dump(run_object)
 
 
+def _is_postgres_integrity(exc):
+    return isinstance(exc, psycopg2.IntegrityError)
+
+
+@retry(wait_exponential_multiplier=1000, wait_exponential_max=10000, retry_on_exception=_is_postgres_integrity)
```
wilko77: this could potentially go on forever. Be a good parent and set a limit.

backend/entityservice/views/run/description.py
hardbyte: Ta didn't know that - have switched

backend/entityservice/views/run/description.py
hardbyte: In the docs it said it would go up to the `wait_exponential_max` seconds. Same with the version of moved to from the tenacity docs

backend/entityservice/views/run/description.py
```diff
@@ -23,21 +25,29 @@ def get(project_id, run_id):
     return RunDescription().dump(run_object)
 
 
+@retry(wait=wait_random_exponential(multiplier=1, max=60), retry=retry_if_exception_type(psycopg2.IntegrityError))
```
wilko77: I still think there should also be a `stop=stop_after...` condition to make sure we not retrying forever.

hardbyte: I verified this - you are correct

PR #422, opened 2019-09-06
Feature export metrics

hardbyte:
Adds an endpoint to the flask app `/api/v1/metrics/` which exposes prometheus metrics.

Note this PR adds a new restriction that gunicorn only uses 1 process, although it should be possible to overcome this within a single pod - but a different approach is required to support multiple flask app pods (which is a shame).

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -34,9 +34,11 @@ USER user
 ADD . /var/www
 
 # Serve using gunicorn. Ideally this has nginx in front of it!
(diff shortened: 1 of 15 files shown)


hardbyte:
I'm leaning towards a side car metrics exporter - it could read from the database (maybe even subscribe to events) and redis to provide richer information without imposing any restrictions on the flask app.

hardbyte:
Closing as stale

PR #424, opened 2019-09-09
Use groups instead of mapping in the benchmark.

gusmith:
First of two PRs related to multi party linkage.
First one: using groups instead of mappings, and introducing a new metric for the grouping output.
In this PR, it is still hard coded to be ran for two parties only.

### DIFF
--- a/benchmarking/benchmark.py
+++ b/benchmarking/benchmark.py
@@ -207,7 +207,6 @@ def score_accuracy(groups, truth_groups):
 
 def compose_result(status, tt, experiment, sizes, threshold):
     positives, negatives, false_positives = tt
-    sizes_dic = {'size_{}'.format(chr(97 + i)): sizes[i] for i in range(len(sizes))}
     # Compute the accuracy, i.e. number of positives divided by the number of matched entities in the ground
     # truth (positives + negatives)
     accuracy = positives / (positives + negatives)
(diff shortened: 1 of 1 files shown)


benchmarking/benchmark.py
```diff
@@ -149,48 +149,77 @@ def upload_data(participant, auth_token, clk_length):
...
+    positives = len(truth_groups_set.intersection(set(transformed_groups)))
+    negatives = len(truth_groups_set) - positives
+    false_positives = len(transformed_groups) - positives
+    return positives, negatives, false_positives
 
 
 def compose_result(status, tt, experiment, sizes, threshold):
-    tp, tn, fp, fn = tt
+    positives, negatives, false_positives = tt
+    sizes_dic = {'size_{}'.format(chr(97 + i)): sizes[i] for i in range(len(sizes))}
```
wilko77: what's that for? It isn't used anywhere. I guess you eventually will build the 'sizes' section in the `result` with that. Am I right?

gusmith:
Sorry Wilko, doing some force-push to rebase on top of latest develop.

benchmarking/benchmark.py
gusmith: You're right. I was trying to split the multi-party linkage benchmark update in two PRs, but this is a remaining of a wrong split. It becomes useful in the following PR.

PR #425, opened 2019-09-09
Benchmark multi party

gusmith:
Second PRs for multi party linkage benchmark.
Based on the group results from the PR #424 , it simply handle more than two parties which was quite hard coded in the benchmark.

### DIFF
--- a/benchmarking/benchmark.py
+++ b/benchmarking/benchmark.py
@@ -147,9 +147,9 @@ def upload_binary_clks(config, sizes, credentials):
         logger.info('uploading clks for {} took {:.3f}'.format(participant, tick() - start))
 
     assert len(sizes) == len(credentials['update_tokens'])
-    for participant in range(len(sizes)):
+    for participant_id in range(len(sizes)):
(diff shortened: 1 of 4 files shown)


benchmarking/benchmark.py
```diff
@@ -144,27 +146,29 @@ def upload_data(participant, auth_token, clk_length):
             logger.warning('oh no...\n{}'.format(e))
         logger.info('uploading clks for {} took {:.3f}'.format(participant, tick() - start))
 
-    upload_data('a', credentials['update_tokens'][0], length_a)
-    upload_data('b', credentials['update_tokens'][1], length_b)
+    assert len(sizes) == len(credentials['update_tokens'])
+    for participant in range(len(sizes)):
```
wilko77: you are re-using a variable name for a variable of a different kind. Before, `participant` was a character, now it is an integer. This might lead to confusion and makes your code harder to read. Please avoid in the future. 
(You could rename this one to `participant_id`)

gusmith: :+1:

PR #428, opened 2019-09-17
Slower tests

gusmith:
Update some tests which were failing.  
This branch is based on `deployment-documentation` which modified the default deployment size.
What surprises me is that the `minimal-values` we were using for the k8s tests were supposedly smaller than the default one on the mentioned branch, but the tests take longer...


gusmith:
Instead of creating this PR on `deployment-documentation`, create a new one on develop.

Issue #421, opened 2019-09-06
celery queue takes too long to be cleaned after deleting a run


gusmith:
### Deployment:
```
 'version': {'anonlink': '0.11.2',
             'entityservice': 'v1.11.2', (dev branch)
             'python': '3.7.3'}
```
Deployed using `docker-compose` on a "smallish" machine (my laptop).
Not setting the option `max-tasks-per-child` to a small value.

### Observation:
While running an experiment for multi party linkage (1M * 1M * 1M), the benchmark timed out as expected, which triggered the run deletion.
The runs is well deleted on the database, but it took around 20 minutes for the workers to empty the celery queue full of tasks meant to fail fast (raising a `DBResourceMissing` as expected). 

20 minutes seem WAY too long to simply remove a task without computing anything. It still requires to query the database, but does not create any computing.

hardbyte:
I think we first need to measure where the tune is being spent.

What options come to mind to speed this up?

- The workers could check redis instead of postgres.
- we could have an aggregation of comparison tasks, e.g., after the first 10k tasks the rest get combined into a single job that checks the job is still valid before scheduling the next 10k...

hardbyte:
Closed in #432

PR #432, opened 2019-09-21
Feature cache runs

hardbyte:
This builds upon  https://github.com/data61/anonlink-entity-service/pull/431 to track run state in the redis cache.

This should address #421

### DIFF
--- a/backend/entityservice/cache/active_runs.py
+++ b/backend/entityservice/cache/active_runs.py
@@ -1,0 +1,25 @@
+import structlog
+
+from entityservice.cache.connection import connect_to_redis
+from entityservice.cache.helpers import _get_run_hash_key
+
+logger = structlog.get_logger()
+
+
+def set_run_state(run_id, state='active'):
+    logger.info("Setting run active", run_id=run_id)
+    r = connect_to_redis()
+    key = _get_run_hash_key(run_id)
+    r.hset(key, 'state', state)
+
+
+def get_run_state(run_id):
+    r = connect_to_redis(read_only=True)
+    logger.info("Getting run state", run_id=run_id)
+    key = _get_run_hash_key(run_id)
+    return r.hget(key, 'state')
+
+
+def is_run_active(run_id):
+    return 'active' == get_run_state(run_id)
+
--- a/backend/entityservice/cache/active_runs.py
+++ b/backend/entityservice/cache/active_runs.py
@@ -1,3 +1,10 @@
+"""
+Cache information about the run state in redis.
+
+Note across the cache we use a common redis hash for each run,
+this module uses the 'state' key.
+"""
+from enum import Enum
 import structlog
 
 from entityservice.cache.connection import connect_to_redis
@@ -6,20 +13,37 @@ from entityservice.cache.helpers import _get_run_hash_key
 logger = structlog.get_logger()
 
 
-def set_run_state(run_id, state='active'):
-    logger.info("Setting run active", run_id=run_id)
+class RunState(Enum):
+    MISSING = None
+    ACTIVE = b'active'
+    DELETED = b'deleted'
+
+
+def _set_run_state(run_id, state):
+    logger.debug(f"Setting run state to '{state.name}'", run_id=run_id)
     r = connect_to_redis()
     key = _get_run_hash_key(run_id)
-    r.hset(key, 'state', state)
+    return r.hset(key, 'state', state.value)
 
 
-def get_run_state(run_id):
+def _get_run_state(run_id):
     r = connect_to_redis(read_only=True)
-    logger.info("Getting run state", run_id=run_id)
     key = _get_run_hash_key(run_id)
-    return r.hget(key, 'state')
+    # hget returns None if missing key/name, and bytes if present
+    maybe_state = r.hget(key, 'state')
+    logger.debug("Loaded run state from cache", run_id=run_id, state=maybe_state)
+    return RunState(maybe_state)
+
+
+def set_run_state_deleted(run_id):
+    return _set_run_state(run_id, state=RunState.DELETED)
+
+
+def set_run_state_active(run_id):
+    return _set_run_state(run_id, state=RunState.ACTIVE)
 
 
 def is_run_active(run_id):
-    return 'active' == get_run_state(run_id)
+    return RunState.ACTIVE == _get_run_state(run_id)
+
 
--- a/backend/entityservice/cache/active_runs.py
+++ b/backend/entityservice/cache/active_runs.py
@@ -57,3 +57,7 @@ def is_run_active(run_id):
     return RunState.ACTIVE == _get_run_state(run_id)
 
 
+def clear_run_state(run_id):
+    r = connect_to_redis()
+    key = _get_run_hash_key(run_id)
+    r.hdel(key, 'state')
--- a/backend/entityservice/cache/active_runs.py
+++ b/backend/entityservice/cache/active_runs.py
@@ -26,6 +26,7 @@ class RunState(Enum):
     """
     MISSING = None
     ACTIVE = b'active'
+    COMPLETE = b'complete'
     DELETED = b'deleted'
 
 
@@ -51,6 +52,10 @@ def set_run_state_deleted(run_id):
 
 def set_run_state_active(run_id):
     return _set_run_state(run_id, state=RunState.ACTIVE)
+
+
+def set_run_state_complete(run_id):
+    return _set_run_state(run_id, state=RunState.COMPLETE)
 
 
 def is_run_active(run_id):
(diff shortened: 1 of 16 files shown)


backend/entityservice/cache/active_runs.py
```diff
@@ -0,0 +1,49 @@
...
+from entityservice.cache.connection import connect_to_redis
+from entityservice.cache.helpers import _get_run_hash_key
+
+logger = structlog.get_logger()
+
+
+class RunState(Enum):
+    MISSING = None
+    ACTIVE = b'active'
+    DELETED = b'deleted'
```
gusmith: Need documentation about these states: what is an active run? Does that mean it is running? What about finished runs?

backend/entityservice/cache/active_runs.py
```diff
@@ -0,0 +1,49 @@
...
+def set_run_state_deleted(run_id):
+    return _set_run_state(run_id, state=RunState.DELETED)
+
+
+def set_run_state_active(run_id):
+    return _set_run_state(run_id, state=RunState.ACTIVE)
+
+
+def is_run_active(run_id):
+    return RunState.ACTIVE == _get_run_state(run_id)
```
gusmith: If the run state is missing from the cache, shouldn't we check the information from the database?
Which leads to another question: how long are we keeping this information in redis? Shouldn't we clean the inactive keys? Should we set an expiry time for the keys?

backend/entityservice/views/run/description.py
```diff
@@ -31,6 +32,7 @@ def get(project_id, run_id):
 def _delete_run(run_id, log):
     # Retry if a db integrity error occurs (e.g. when a worker is writing to the same row)
     # Randomly wait up to 2^x * 1 seconds between each retry until the range reaches 60 seconds, then randomly up to 60 seconds afterwards
+    set_run_state_deleted(run_id)
```
gusmith: Are we doing anything different if the run is missing or deleted? If no, could we simply delete the key from redis?

backend/entityservice/cache/active_runs.py
hardbyte: I'm happy for redis to be the source of truth - after all the entire job is screwed if we lose our celery queue which is stored in Redis. In any case such a check wouldn't be here in the `cache.py` file.

Sure we should we delete keys from redis at some point, not entirely sure it make sense now though. Given we are talking about a handful of bytes per run and by default we provision redis-ha with 1.5GiB of memory... I don't see any great urgency. Would you be happy if I put in an expiry time of a 10 days?

backend/entityservice/cache/active_runs.py
hardbyte: Sure I'll add a docstring but I'll point out this isn't exposed outside of this file - it is simply to support the public api of: `is_run_active`, `set_run_state_active` and `set_run_state_deleted`.

We could add (and expose) states like "finished" - if and when required.

backend/entityservice/views/run/description.py
hardbyte: A missing run means something has gone wrong where a deletion is normal behavior. 

I could be convinced but I'd prefer to set the state in redis so the backend tasks can act accordingly - perhaps throw a big tantrum if the run is missing, but just exit gracefully if the run has been deleted.

backend/entityservice/cache/active_runs.py
hardbyte: I will however add a `clear_run_state` function to align with the `entityservice.cache.progress` api.

backend/entityservice/cache/active_runs.py
gusmith: I was also wondering yesterday, should we include an `ERROR` state?

backend/entityservice/cache/active_runs.py
gusmith: A default of 10 days or more is perfect (configurable even better :) ).
It's mainly to start having best practices when using redis.

gusmith:
Also thinking that we should have at one point some documentation about the different key values stored in redis.

backend/entityservice/cache/active_runs.py
hardbyte: Perhaps at some point, but I don't want to prematurely add things to the cache that are not used.

deployment/entity-service/templates/configmap.yaml
```diff
@@ -59,9 +59,7 @@ data:
   CHUNK_SIZE_AIM: {{ required "workers.CHUNK_SIZE_AIM is required." .Values.workers.CHUNK_SIZE_AIM | quote }}
 
   MAX_CACHE_SIZE: {{ required "workers.MAX_CACHE_SIZE is required." .Values.workers.MAX_CACHE_SIZE | quote }}
-
-  # ENTITY_MATCH_THRESHOLD not provided
```
gusmith: Why did you delete this two lines? The two variables are still in use in the file `settings.py`.

backend/entityservice/settings.py
```diff
@@ -52,7 +53,10 @@ class Config(object):
     CELERY_RESULT_BACKEND_TRANSPORT_OPTIONS = {'master_name': "mymaster"} if REDIS_USE_SENTINEL else {}
 
     CELERY_ANNOTATIONS = {
-        'async_worker.calculate_mapping': {'rate_limit': '1/s'}
+        # Note that these are per worker instance per task rate limits.
```
gusmith: Is that why it was taking forever to empty the queue?

docs/development.rst
```diff
@@ -32,6 +32,35 @@ variables. Multiple workers can be used to distribute the work beyond
...
+Redis Cache: Run Info
+~~~~~~~~~~~~~~~~~~~~~
+
+The run info ``HASH`` stores:
+
+- similarity scoring progress for each run under ``"progress"``
+- run state under ``"state"``, current valid states are
+  ``{active, complete, deleted}``. See
+  ``backend/entityservice/cache/active_runs.py`` for implementation.
+
```
gusmith: :+1:

backend/entityservice/tasks/run.py
```diff
@@ -25,22 +26,26 @@ def prerun_check(project_id, run_id, parent_span=None):
...
             return
 
-        if state in {'running', 'completed', 'error'}:
+        if db_state in {'running', 'completed', 'error'}:
             log.warning("Run already started. Skipping")
             return
 
-        log.debug("Setting run as in progress")
+        # being very defensive here checking if the run state is already in the redis cache
+        if not is_run_missing(run_id):
```
gusmith: Small comment: shouldn't we first check the cache and then the db check? Like that if the check fail with the cache, we don't connect to the db which takes longer. But not important as it is used a single time per run.

backend/entityservice/utils.py
```diff
@@ -124,16 +124,17 @@ def generate_code(length=24):
...
             parties_contributed = get_number_parties_ready(conn, project_id)
-            logger.info("Parties where data is ready: {}".format(parties_contributed))
+            log.info("Parties where data is ready: {}".format(parties_contributed))
         else:
             parties_contributed = get_number_parties_uploaded(conn, project_id)
-            logger.info("Parties where data is uploaded: {}".format(parties_contributed))
+            log.info("Parties where data is uploaded: {}".format(parties_contributed))
         number_parties = get_project_column(conn, project_id, 'parties')
-    logger.info("{}/{} parties have contributed clks".format(parties_contributed, number_parties))
+    log.info("{}/{} parties have contributed clks".format(parties_contributed, number_parties))
```
gusmith: Quick question: should we open an issue to be sure that all methods using logging are using project and run id when available?

deployment/entity-service/templates/configmap.yaml
hardbyte: This setting hasn't had any effect for a long time as the `threshold` is required when posting a new run - see spec [here](https://anonlink-entity-service.readthedocs.io/en/latest/api.html#operation/entityservice.views.run.list.post) - I'll create an issue (https://github.com/data61/anonlink-entity-service/issues/443) and aim to  remove it from settings too.

backend/entityservice/settings.py
hardbyte: No I don't think so because the `calculate_mapping` task was the only one rate limited and there actually isn't any task called that any more - hence the small correction to a correct path but commented out.

backend/entityservice/tasks/run.py
hardbyte: Fair point, I'll reorder.

I was actually wondering about using a redis lock for this whole section but didn't want to introduce it here.

backend/entityservice/utils.py
hardbyte: Sure, I just try improve this kind of thing as I go

PR #433, opened 2019-09-24
Benchmark test on CI runs 3 parties linkage 

gusmith:
Threshold 0.8 is chosen for the 3 party case because the results were quite good.

### DIFF
--- a/benchmarking/linkage-bench-cache-experiments.json
+++ b/benchmarking/linkage-bench-cache-experiments.json
@@ -2,5 +2,9 @@
   {
     "sizes": ["100K", "100K"],
     "threshold": 0.95
+  },
+  {
+    "sizes": ["10K", "10K", "10K"],
+    "threshold": 0.8
   }
 ]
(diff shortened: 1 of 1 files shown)


gusmith:
On my machine, results from running the benchmark (remove all the unnecessary values), the provided json may not be valid as I'm removing by hand the values:
```
{
  'sizes': ['100K', '100K'],
  'threshold': 0.95,
  'groups_results': {
     'accuracy': 0.43175537666706715,
     'false_positives': 0,
     'negatives': 28377,
     'positives': 21561}
  'timings': {'runtime': 27.936643},
},
{
  'sizes': ['100K', '100K'],
  'threshold': 0.8,
  'groups_results': {
    'accuracy': 0.9962553566422364,
    'false_positives': 7289,
    'negatives': 187,
    'positives': 49751
  },
  'timings': {'runtime': 38.349424}
}
```
So running the 2 parties 100K experiment improves drastically the accuracy using a threshold of 0.8 instead of 0.95, but increases runtime by 10 seconds. So I'll include the change in this PR too.

PR #436, opened 2019-09-25
Make sentinel name configurable

hardbyte:
As Wilko pointed out in https://github.com/data61/anonlink-entity-service/pull/430#discussion_r327941353 all the other redis settings are fully configurable except for the sentinel name. This fixes this poor state of affairs.

### DIFF
--- a/backend/entityservice/cache/connection.py
+++ b/backend/entityservice/cache/connection.py
@@ -7,7 +7,7 @@ from entityservice.settings import Config as config
 logger = structlog.get_logger()
 redis_host = config.REDIS_SERVER
 redis_pass = config.REDIS_PASSWORD
-redis_port = 26379
+redis_sentinel_port = 26379
 
 
 def connect_to_redis(read_only=False):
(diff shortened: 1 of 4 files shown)


backend/entityservice/cache/connection.py
```diff
@@ -17,13 +17,14 @@ def connect_to_redis(read_only=False):
...
+        sentinel_service = Sentinel([(redis_host, redis_port)], password=redis_pass, socket_timeout=5)
+        sentinel_name = config.REDIS_SENTINEL_NAME
         if read_only:
             logger.debug("Looking up read only redis slave using sentinel protocol")
-            r = sentinel.slave_for('mymaster')
+            r = sentinel_service.slave_for(sentinel_name)
         else:
             logger.debug("Looking up redis master using sentinel protocol")
-            r = sentinel.master_for('mymaster')
+            r = sentinel_service.master_for('mymaster')
```
wilko77: you would have to replace it here as well...

PR #438, opened 2019-09-25
Run integration tests in parallel

hardbyte:
Old:

image

New:

image


hardbyte:
Good point, I was probably a bit too excited about the faster feedback and asked for a review before writing up PR.

I deleted the one test that specifically looked for side effects while failing to add a project - I couldn't think of a better way (under our current api only test strategy) of rewriting that test so I deleted it.

With our hundreds of tests running in parallel we need to increase the rate limits on the service. I noticed it wasn't configurable hence https://github.com/data61/anonlink-entity-service/issues/440
But I don't think there is an issue increasing it even another order of magnitude because the ingress would be the more logical place in a production deployment for service/deployment specific rate limits. Our nginx app simply wants to avoid a DDOS against our gunicorn/python/flask app.

PR #444, opened 2019-09-30
Remove unused setting ENTITY_MATCH_THRESHOLD

### LINKED ISSUE #443: remove deployment wide threshold
There is an environment variable `ENTITY_MATCH_THRESHOLD` which has no effect as creating a run requires a threshold. This issue is to remove the old setting entirely.

hardbyte:
Closes #443

See discussion in https://github.com/data61/anonlink-entity-service/pull/432/#discussion_r329404684

### DIFF
--- a/backend/entityservice/models/run.py
+++ b/backend/entityservice/models/run.py
@@ -3,7 +3,6 @@ from structlog import get_logger
 from entityservice.cache import progress as progress_cache
 import entityservice.database as db
 from entityservice.database import DBConn
-from entityservice.settings import Config as config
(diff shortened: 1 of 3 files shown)


backend/entityservice/models/run.py
```diff
@@ -73,7 +73,7 @@ def __init__(self, project_id, threshold, name, notes):
     @staticmethod
     def from_json(data, project_id):
         # Get optional fields from JSON data
```
gusmith: Should modify the comment: the `threshold` is not optional, as specified in the swagger file.

Issue #418, opened 2019-09-05
Azure Devops CI isn't saving container logs


hardbyte:
This makes debugging intermittent test failures such as race conditions very difficult!

gusmith:
Closed by #445

PR #445, opened 2019-10-01
Feature k8s pods log extraction

gusmith:
Add an extra step in our azure pipeline to publish the logs of the entity-service pods when tests are failing.
First, use a python script to download the logs (easier with python than bash, and I already had the script locally), then, create an azure template for steps running the python script and publishing the logs in azure artifacts, finally integrate this template in the main pipeline.

Note: I included a commit which will need to be reversed: I introduce a test made to fail (directly raising an exception).

### DIFF
--- a/.azurePipeline/getPodsLogs.py
+++ b/.azurePipeline/getPodsLogs.py
@@ -44,8 +44,9 @@ def info_from_pod(_pod, _directory, _release_name):
                 'name': name_container, 'image': image_name, 'restart_count': restart_count}
         
         with open(_directory / 'pod-{}_container-{}.log'.format(info['short_pod_name'], info['name']), 'wt') as f:
-            json.dump(info, f, indent=2, sort_keys=True)
-            f.write('\n\n')
+            for key, value in info.items():
+                f.write("{}: {}\n".format(key, value))
+            f.write('{}\n\n'.format('-'*25))
         with open(_directory / 'pod-{}_container-{}.log'.format(info['short_pod_name'], info['name']), 'at') as f:
             get_logs_container(info, f)
         if restart_count > 0:
--- a/.azurePipeline/getPodsLogs.py
+++ b/.azurePipeline/getPodsLogs.py
@@ -32,14 +32,15 @@ def get_logs_container(container_info, file, previous=False):
     file.write(colors.strip_color(output.stdout.decode("utf-8")))
 
 
-def info_from_pod(_pod, _directory):
+def info_from_pod(_pod, _directory, _release_name):
     pod_name = _pod.get('metadata').get('name')
     list_containers = _pod.get('status').get('containerStatuses')
     for container in list_containers:
         name_container = container.get('name')
         restart_count = container.get('restartCount')
         image_name = container.get('image')
-        info = {'full_pod_name': pod_name, 'short_pod_name': pod_name[len('benchmark-es-data61-xyz-'):],
+        prefix_to_remove = "{}-".format(_release_name)
+        info = {'full_pod_name': pod_name, 'short_pod_name': pod_name[len(prefix_to_remove):],
                 'name': name_container, 'image': image_name, 'restart_count': restart_count}
         
         with open(_directory / 'pod-{}_container-{}.log'.format(info['short_pod_name'], info['name']), 'wt') as f:
@@ -66,7 +67,7 @@ def main():
     release_name = sys.argv[2]
     list_pods = get_list_pods(release_name)
     for pod in list_pods:
-        info_from_pod(pod, directory)
+        info_from_pod(pod, directory, release_name)
 
 
 if __name__ == '__main__':
--- a/.azurePipeline/getPodsLogs.py
+++ b/.azurePipeline/getPodsLogs.py
@@ -1,0 +1,70 @@
+#!/usr/bin/env python3
+
+"""
+The script will currently be used in the Azure Pipeline to get back the logs of the pods when a test is failing
+allowing developers to document the issue.
+"""
+
+import json
+from pathlib import Path
+import subprocess
+import sys
+
+# The namespace is not configurable as we are expected to run this script only in Azure Dev-Ops which deployed the
+# service only in this namespace.
+NAMESPACE = "test-azure"
+
+
+def get_list_pods(_release_name):
+    cmd = "kubectl --namespace {} get pods -l release={} -o json".format(NAMESPACE, _release_name)
+    r = subprocess.run(cmd.split(" "), capture_output=True)
+    items_from_release = json.loads(r.stdout)['items']
+    return items_from_release
+
+
+def get_logs_container(container_info, file, previous=False):
+    cmd = "kubectl --namespace {} logs {} {}".format(NAMESPACE, container_info['full_pod_name'], container_info['name'])
+    if previous:
+        cmd += " --previous"
+    subprocess.call(cmd.split(" "), stdout=file)
+
+
+def info_from_pod(_pod, _directory):
+    pod_name = _pod.get('metadata').get('name')
+    list_containers = _pod.get('status').get('containerStatuses')
+    for container in list_containers:
+        name_container = container.get('name')
+        restart_count = container.get('restartCount')
+        image_name = container.get('image')
+        info = {'full_pod_name': pod_name, 'short_pod_name': pod_name[len('benchmark-es-data61-xyz-'):],
+                'name': name_container, 'image': image_name, 'restart_count': restart_count}
+        
+        with open(_directory / 'pod-{}_container-{}.log'.format(info['short_pod_name'], info['name']), 'wt') as f:
+            json.dump(info, f, indent=2, sort_keys=True)
+            f.write('\n\n')
+        with open(_directory / 'pod-{}_container-{}.log'.format(info['short_pod_name'], info['name']), 'at') as f:
+            get_logs_container(info, f)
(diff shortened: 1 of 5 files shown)


gusmith:
see https://dev.azure.com/data61/Anonlink/_build/results?buildId=614&view=results for a failed build publishing the pods logs.

.azurePipeline/getPodsLogs.py
```diff
@@ -0,0 +1,70 @@
...
+        restart_count = container.get('restartCount')
+        image_name = container.get('image')
+        info = {'full_pod_name': pod_name, 'short_pod_name': pod_name[len('benchmark-es-data61-xyz-'):],
+                'name': name_container, 'image': image_name, 'restart_count': restart_count}
+        
+        with open(_directory / 'pod-{}_container-{}.log'.format(info['short_pod_name'], info['name']), 'wt') as f:
+            json.dump(info, f, indent=2, sort_keys=True)
+            f.write('\n\n')
+        with open(_directory / 'pod-{}_container-{}.log'.format(info['short_pod_name'], info['name']), 'at') as f:
+            get_logs_container(info, f)
```
hardbyte: Mixing json and logging output is a bit strange

.azurePipeline/getPodsLogs.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def info_from_pod(_pod, _directory):
+    pod_name = _pod.get('metadata').get('name')
+    list_containers = _pod.get('status').get('containerStatuses')
+    for container in list_containers:
+        name_container = container.get('name')
+        restart_count = container.get('restartCount')
+        image_name = container.get('image')
+        info = {'full_pod_name': pod_name, 'short_pod_name': pod_name[len('benchmark-es-data61-xyz-'):],
```
hardbyte: `benchmark-es-data61-xyz-`?

.azurePipeline/getPodsLogs.py
```diff
@@ -0,0 +1,70 @@
...
+import subprocess
+import sys
+
+# The namespace is not configurable as we are expected to run this script only in Azure Dev-Ops which deployed the
+# service only in this namespace.
+NAMESPACE = "test-azure"
+
+
+def get_list_pods(_release_name):
+    cmd = "kubectl --namespace {} get pods -l release={} -o json".format(NAMESPACE, _release_name)
```
hardbyte: Did you consider installing the python-kubernetes package and using that instead of shelling out?

https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/CoreV1Api.md#read_namespaced_pod_log

.azurePipeline/templatePublishLogsFromPods.yml
```diff
@@ -0,0 +1,30 @@
...
+  displayName: 'Copy the logs from the service pods.'
+  condition: failed()
+  env:
+    KUBECONFIG: ${{ parameters.kubeConfigFile }}
+- task: PublishPipelineArtifact@1
+  displayName: 'Publish pods logs in Azure'
+  inputs:
+    artifactName: PodLogs
+    targetPath: ${{ parameters.logsFolderPath }}
+  condition: failed()
```
hardbyte: nit: newline

.azurePipeline/templatePublishLogsFromPods.yml
```diff
@@ -0,0 +1,30 @@
...
+parameters:
+  logsFolderPath: ''
+  releaseName: ''
+  kubeConfigFile: ''
+
+steps:
+- task: UsePythonVersion@0
+  inputs:
+    versionSpec: '3.7'
+  condition: failed()
```
hardbyte: I suggest we use Azure's variables feature here so we can decide in the ci/cd system to keep the logs without having to inject a test failure.

https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch

.azurePipeline/templatePublishLogsFromPods.yml
```diff
@@ -0,0 +1,30 @@
...
+- script: |
+    echo "Copy the pods' logs from the release '${{ parameters.releaseName }}' to '${{ parameters.logsFolderPath }}'."
+    python .azurePipeline/getPodsLogs.py ${{ parameters.logsFolderPath }} ${{ parameters.releaseName }}
+  displayName: 'Copy the logs from the service pods.'
+  condition: failed()
```
hardbyte: Do we have to repeat the condition in this template, can we lift it up to the calling location in `azure-pipelines.yml`?

.azurePipeline/getPodsLogs.py
```diff
@@ -0,0 +1,70 @@
...
+    r = subprocess.run(cmd.split(" "), capture_output=True)
+    items_from_release = json.loads(r.stdout)['items']
+    return items_from_release
+
+
+def get_logs_container(container_info, file, previous=False):
+    cmd = "kubectl --namespace {} logs {} {}".format(NAMESPACE, container_info['full_pod_name'], container_info['name'])
```
hardbyte: Note this command gets logs with ascii color codes... Is this what you want?

.azurePipeline/templatePublishLogsFromPods.yml
```diff
@@ -0,0 +1,30 @@
...
+  env:
+    KUBECONFIG: ${{ parameters.kubeConfigFile }}
+- task: PublishPipelineArtifact@1
```
hardbyte: I'm guessing you didn't want to just `cat` them all into the Azure Devops output log? It would make it even easier to diagnose a test failure...

.azurePipeline/getPodsLogs.py
gusmith: That explains why the short pod name was quite small... :+1:

.azurePipeline/getPodsLogs.py
gusmith: But the information is relevant.
From 
```
{
  "full_pod_name": "esdf764b0f-memstore-server-0",
  "image": "redis:4.0.11-alpine",
  "name": "redis",
  "restart_count": 0,
  "short_pod_name": "er-0"
}

[logs]
```
would
```
"full_pod_name": "esdf764b0f-memstore-server-0",
"image": "redis:4.0.11-alpine",
"name": "redis",
"restart_count": 0,
"short_pod_name": "er-0"
----------------------------------------------------------------

[logs]
```
be better?
The most interesting from it it the image and the restart count. The other information are more relevant if the deployment was staying up.

gusmith:
It will close https://github.com/data61/anonlink-entity-service/issues/418

.azurePipeline/getPodsLogs.py
gusmith: I didn't even look if there was a package as I'm really used to the shell commands.
But I'll open an issue to record it (#446 ). To be honest, I also like the fact that the script does not have any requirement.

.azurePipeline/templatePublishLogsFromPods.yml
gusmith: The step `template` does not accept a `condition` field, which is why I needed to bring them back here.

.azurePipeline/templatePublishLogsFromPods.yml
gusmith: I do not see the point of using variables for this. In my mind, there three options:
 - we should publish the logs for every build (changing the condition from `failed()` to `always()`
 - we should never publish the logs (which I would assume is not preferred, otherwise we wouldn't have opened the issue #418 )
 - the logs are useful in some cases.
I went with the third option as I do not think that they are useful when a build succeeds (but I'm not feeling to strongly about it).
However, adding a variable `true` or `false` is in my opinion a bad idea, as I do not believe a developer should have to modify the CI script on a per branch basis.

.azurePipeline/templatePublishLogsFromPods.yml
gusmith: :+1:

.azurePipeline/templatePublishLogsFromPods.yml
gusmith: Compromise: the logs are always provided.

.azurePipeline/getPodsLogs.py
gusmith: Use [ansioclors](https://pypi.org/project/ansicolors/) to remove colors from logs.

.azurePipeline/templatePublishLogsFromPods.yml
gusmith: I don't really like looking at logs from the Azure Dev-ops page, which is why I tend to prefer files I can view with the tools I prefer. If people prefer from the Azure dev-ops page, we can indeed `cat` the files, but will need to choose a preferred way to split the output in a nice way.

.azurePipeline/getPodsLogs.py
hardbyte: Sure the info is great. It was more of an observation, no need to change it.

.azurePipeline/getPodsLogs.py
hardbyte: Okay, of course if we are outputting to a console the colors are really good to keep. All good by me either way

.azurePipeline/templatePublishLogsFromPods.yml
hardbyte: Fair enough!

PR #448, opened 2019-10-09
Feature k8s logging

gusmith:
This PR is about adding possibility to configure the entity-service logging in deployment.
To do this, a user has to:
 - create a logging configuration file which the entity-service can use (out of scope for this PR)
 - create and publish on k8s a configmap from this logging configuration file (documented)
 - add the corresponding parameters in the deployment configuration (documented in docs and in `values.yaml`)

Under the hood and following the documentation, a configmap is created from the logging file with a selected key. Without a key, each key from the configmap would create its own file when loaded in a volume.
Then, when selecting a logging configmap at deployment time, the env var configmap will update the env var `LOG_CFG` pointing to a file which will be created from the logging configmap and mounted to each container which uses the env var `LOG_CFG`, i.e. the workers, the `init_db`, and the entity flask.

Some notes:
 - I would have loved another solution where you can simply provide a path to a file when starting a deployment, but this is not yet by kubernetes (it cannot load a file which the templates are not aware of, which is not convenient for users)
 - providing the whole log file directly in the deployment configuration is a bad idea because some characters are interpreted by kubernetes (mainly the %, but maybe others)
 - providing a default configuration in the `vlaues.yaml` file is a bad idea because a user would have to overwrite each single value, making it a horrible experience to remove a section of the default configuration (as we need to overwrite each value, nothing is really deleted from the defaults)
 - this logging config map could be extended to include more generic configurations read from files if required.

### DIFF
--- a/deployment/entity-service/templates/api-deployment.yaml
+++ b/deployment/entity-service/templates/api-deployment.yaml
@@ -94,18 +94,18 @@ spec:
             initialDelaySeconds: 60
             periodSeconds: 60
             timeoutSeconds: 5
-          {{- if .Values.workers.loggingCfgm }}
+          {{- if .Values.workers.loggingCfg }}
           volumeMounts:
             - name: config-volume
               mountPath: /var/config
           {{- end }}
-      {{- if .Values.workers.loggingCfgm }}
+      {{- if .Values.workers.loggingCfg }}
       volumes:
         - name: config-volume
           configMap:
-            name: {{ required ".workers.logging_cfgm.name is required to provide the name of the configmap where the logging configuration is stored." .Values.workers.loggingCfgm.name }}
+            name: {{ template "es.fullname" . }}-logging
             items:
-              - key: {{ required ".workers.logging_cfgm.key is required to provide the key in the configmap which value is the content of the logging configuration." .Values.workers.loggingCfgm.key }}
+              - key: "loggingCfg"
                 path: "logging_file.yaml"
       {{- end }}
       {{- if .Values.api.pullSecret }}
--- a/deployment/entity-service/templates/api-deployment.yaml
+++ b/deployment/entity-service/templates/api-deployment.yaml
@@ -94,6 +94,20 @@ spec:
             initialDelaySeconds: 60
             periodSeconds: 60
             timeoutSeconds: 5
+          {{- if .Values.workers.loggingCfgm }}
+          volumeMounts:
+            - name: config-volume
+              mountPath: /var/config
+          {{- end }}
+      {{- if .Values.workers.loggingCfgm }}
+      volumes:
+        - name: config-volume
+          configMap:
+            name: {{ required ".workers.logging_cfgm.name is required to provide the name of the configmap where the logging configuration is stored." .Values.workers.loggingCfgm.name }}
+            items:
+              - key: {{ required ".workers.logging_cfgm.key is required to provide the key in the configmap which value is the content of the logging configuration." .Values.workers.loggingCfgm.key }}
+                path: "logging_file.yaml"
+      {{- end }}
       {{- if .Values.api.pullSecret }}
       imagePullSecrets:
         - name: {{ .Values.api.pullSecret }}
--- a/deployment/entity-service/templates/api-deployment.yaml
+++ b/deployment/entity-service/templates/api-deployment.yaml
@@ -94,12 +94,12 @@ spec:
             initialDelaySeconds: 60
             periodSeconds: 60
             timeoutSeconds: 5
-          {{- if .Values.workers.loggingCfg }}
+          {{- if .Values.loggingCfg }}
           volumeMounts:
             - name: config-volume
               mountPath: /var/config
           {{- end }}
-      {{- if .Values.workers.loggingCfg }}
+      {{- if .Values.loggingCfg }}
       volumes:
         - name: config-volume
           configMap:
--- a/deployment/entity-service/Chart.yaml
+++ b/deployment/entity-service/Chart.yaml
@@ -1,6 +1,6 @@
 name: entity-service
 appVersion: 1.11.2
-version: 1.12.0
+version: 1.13.0
 description: Privacy preserving record linkage service
 sources:
   - https://github.com/data61/anonlink-entity-service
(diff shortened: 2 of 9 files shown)


hardbyte:
> - providing the whole log file directly in the deployment configuration is a bad idea because some characters are interpreted by kubernetes (mainly the %, but maybe others)
I'm not sure I buy this. I looked at a few example helm charts and found plenty that use `%` - [here is elasticsearch](https://github.com/helm/charts/blob/master/stable/elasticsearch/values.yaml#L97) specifically configuring logging.

See the [redis config map](https://github.com/helm/charts/blob/master/stable/redis-ha/templates/redis-ha-configmap.yaml), or [jenkins](https://github.com/helm/charts/blob/master/stable/jenkins/templates/jenkins-master-backendconfig.yaml) which both just dump the values to a "file" key.

Another approach is taken by [prometheus](https://github.com/helm/charts/blob/master/stable/prometheus/templates/server-configmap.yaml) which assembles a final config map from different sections in the `values.yaml` file.

> - providing a default configuration in the vlaues.yaml file is a bad idea because a user would have to overwrite each single value, making it a horrible experience to remove a section of the default configuration (as we need to overwrite each value, nothing is really deleted from the defaults)

I'm not sure I understand this - are you meaning a horrible experience for us (the service developers) when we change defaults or a horrible experience for someone deploying anonlink-entity-service (if so why?)

gusmith:
The main issue is usability for a user deploying the entity-service who would like to use their own logging configuration.
Providing a default logging via the configmap without having to create beforehand the config map is simple. But being able to provide a chosen one *by the user* is not possible (cf https://helm.sh/docs/chart_template_guide/#lines 

> Currently, there is no way to pass files external to the chart during helm install.

).
But it seems (I just found it, someone found a "solution": https://github.com/helm/helm/issues/3276 (one of the last comment). it seems a bit ugly but could work.

Looking at the redis config map, it's all good as long as not too configurable. The following example will at the same time explain the second quote you used. If I had a default configmap with the data
```
handlers:
  stdout:
    class: logging.StreamHandler
    level: DEBUG
    formatter: simple
    filters: [stdout_filter]
    stream: ext://sys.stdout

  stderr:
    class: logging.StreamHandler
    level: ERROR
    formatter: simple
    filters: [stderr_filter]
    stream: ext://sys.stderr

  info_file_handler:
    class: logging.handlers.RotatingFileHandler
    level: INFO
    formatter: file
    filename: info.log
    maxBytes: 10485760 # 10MB
    backupCount: 20
    encoding: utf8
[...]
```

now, if I want to change one value, it would be easy with a `set` or in a value file, but if want to delete a section, it starts to become more complicated. Let suppose I just want to remove the given level to the handler `stderr`, because in helm a value file can only overwrite values, I would have to set `handlers.stderr.level=''` or `handlers.stderr.level={}` or something else but I do not know exactly which one python would be happy with. So in a way, a user will not be able to provide her logging configuration, but will have to overwrite our defaults (without being sure how the logging system will interpret them).

Finally, providing the whole logging into the yaml file created some issues as some characters were interpreted by helm (the % which is for example used in `format: "%(asctime)-15s %(name)-12s %(levelname)-8s: %(message)s"` is not making helm happy.

hardbyte:
Thanks for providing the extra detail Guillaume - I'm still thinking it would be better to embed the entire logging configuration file into a single configmap key, but I'm open to convincing. 

I'm assuming we can either escape the `%` easily.

> now, if I want to change one value, it would be easy with a set or in a value file, but if want to delete a section, it starts to become more complicated. Let suppose I just want to remove the given level to the handler stderr, because in helm a value file can only overwrite values, I would have to set handlers.stderr.level='' or handlers.stderr.level={} or something else but I do not know exactly which one python would be happy with. So in a way, a user will not be able to provide her logging configuration, but will have to overwrite our defaults (without being sure how the logging system will interpret them).

If the user (person deploying anonlink-entity-service) is writing their own `values.yaml` file wouldn't they just delete the relevant line (`level: ERROR`) in your example?

You say the "redis approach is all good as long as it isn't too configurable" - but as far as I could tell it is completely configurable because the entire file _can be_ provided by the user. In the redis-ha chart's [values.yaml](https://github.com/helm/charts/blob/master/stable/redis-ha/values.yaml#L115):

```
  ## Custom redis.conf files used to override default settings. If this file is
  ## specified then the redis.config above will be ignored.
  # customConfig: |-
      # Define configuration here
```

Which makes it way into the configmap via an include "config-redis.conf":
```
data:
  redis.conf: |
{{- include "config-redis.conf" . }}
```

Which is hiding a bit of complexity for the default case, but if the user has provided a custom config directly in their deployment it just gets rendered straight into the include ([include source file](https://github.com/helm/charts/blob/master/stable/redis-ha/templates/_configs.tpl)):

```
{{- define "config-redis.conf" }}
{{- if .Values.redis.customConfig }}
{{ tpl .Values.redis.customConfig . | indent 4 }}
{{- else }}
... other complex stuff for when the user doesn't provide a whole file here...
```

In any case is there any advantage of having the `values.yaml` expose anything other than the whole logging config file? I'm reluctant to add an extra installation step unless you're confident it is the right approach.

gusmith:
Using `|-` was indeed escaping the special characters.
so with the latest change, if a user provide the key `workers.loggingCfg` in their value file, containing all the configuration file, it will then create a new configmap containing this multi lines string, and adding an environment variable to the entitty-service configmap (where the containers will find the future file). Each container which uses it, will mount a volume built from this logging configmap and will get the logging file accessible at the path provided in the environment variable.

Note that I created a new configmap instead of re-using the entity-service configmap  because the containers are using it to get environment variables, so each key value pair become an environment variable. The current api does not allow to select keys from the configmap, it grabs all of them.
Note 2: I am currently unsure if a container is restarted when a value in any configmap is updated. Which makes me even less sure that updating the logging configmap would restart a container (as it uses it indirectly via a volume).

deployment/entity-service/templates/api-deployment.yaml
```diff
@@ -94,6 +94,20 @@ spec:
             initialDelaySeconds: 60
             periodSeconds: 60
             timeoutSeconds: 5
+          {{- if .Values.workers.loggingCfg }}
```
hardbyte: Suggest removing `workers` from the path as this logging configuration is for the whole service

PR #450, opened 2019-10-10
Issue 447

gusmith:
I think we are often observing deployment slow-starts fails instead of real test failures.
But they fail the builds so let's see what we can do.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -189,17 +189,19 @@ stages:
         --set api.www.image.tag=$(DOCKER_TAG) \
         --set api.app.image.tag=$(DOCKER_TAG) \
         --set api.dbinit.image.tag=$(DOCKER_TAG) \
-        --set workers.image.tag=$(DOCKER_TAG)
+        --set workers.image.tag=$(DOCKER_TAG) \
+        --set workers.replicaCount=4
 
         echo "Dry run"
-        helm upgrade --install --dry-run --wait --namespace $(NAMESPACE) $(DEPLOYMENT) . \
+        helm upgrade --install --debug --dry-run --wait --namespace $(NAMESPACE) $(DEPLOYMENT) . \
         --set api.app.debug=true \
         --set global.postgresql.postgresqlPassword=notaproductionpassword \
         --set api.ingress.enabled=false \
         --set api.www.image.tag=$(DOCKER_TAG) \
         --set api.app.image.tag=$(DOCKER_TAG) \
         --set api.dbinit.image.tag=$(DOCKER_TAG) \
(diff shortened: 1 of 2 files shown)


gusmith:
It has ran 5 times in a row without failure, a 6th is in progress.
So the issue was about accessing an aws bucket: as the tests are ran in parallel, a number of them were trying to create the bucket at the same time, this returning an error when it happens.

azure-pipelines.yml
```diff
@@ -189,17 +189,19 @@ stages:
         --set api.www.image.tag=$(DOCKER_TAG) \
         --set api.app.image.tag=$(DOCKER_TAG) \
         --set api.dbinit.image.tag=$(DOCKER_TAG) \
-        --set workers.image.tag=$(DOCKER_TAG)
+        --set workers.image.tag=$(DOCKER_TAG) \
+        --set workers.replicaCount=4
 
         echo "Dry run"
-        helm upgrade --install --dry-run --wait --namespace $(NAMESPACE) $(DEPLOYMENT) . \
+        helm upgrade --install --debug --dry-run --wait --namespace $(NAMESPACE) $(DEPLOYMENT) . \
```
wilko77: do we really need the debug output?

gusmith: While not being relevant for this PR, here is a `dry-run` output without `debug`:
```
Dry run
Release "esae82d157" does not exist. Installing it now.
NAME:   esae82d157
```
With `debug`:
```
Dry run
[debug] Created tunnel using local port: '42801'

[debug] SERVER: "127.0.0.1:42801"
        volumeMounts:
        - mountPath: /data
          name: data
      - name: sentinel
        image: redis:4.0.11-alpine
        imagePullPolicy: IfNotPresent
        command:
          - redis-sentinel
        args:
          - /data/conf/sentinel.conf
        livenessProbe:
          exec:
            command: ["redis-cli", "-p", "26379", "ping"]
          initialDelaySeconds: 15
          periodSeconds: 5
        readinessProbe:
          exec:
            command: ["redis-cli", "-p", "26379", "ping"]
          initialDelaySeconds: 15
          periodSeconds: 5
        resources:
          limits:
            memory: 256Mi
          requests:
            cpu: 100m
            memory: 128Mi
          
        ports:
          - name: sentinel
            containerPort: 26379
        volumeMounts:
        - mountPath: /data
          name: data
      volumes:
      - name: config
        configMap:
          name: es3b8538bc-memstore-configmap
      - name: data
        emptyDir: {}
Finishing: Lint and dry-run
```
Providing maybe too much details, but without it, the output is also not very useful.

Issue #457, opened 2019-10-31
tests `test_project_run_status` are currently useless (or redundant)


gusmith:
In dev, the tests in `test_project_run_status` are checking if the run is progressing "normally" (i.e. the stages are increasing, the relative progress is also increasing, etc).
But, they all call the method `wait_approx_run_time` which supposedly waits for the time to complete (if the number of clks computed per second is valid), *and* it waits for an extra 5 seconds (which I'm assuming was added because the CI was slow or something).

But each runs tested by this method takes around 1.5s on my machine, so by waiting 5 seconds, we are actually always checking that the run has completed. As the tests are running something like 16 times, we are adding a bit more than an extra minute to the tests for no relevant output as we are already testing these runs to completion.

@wilko77 mentioned that this may need to be integrated in the method `wait_for_run_completion` which could validate every time that the run is progressing. It would be quite nice :)

hardbyte:
@wilko77 you had thought on this regarding removing the tests and adding asserts to all long running tests that ensure the messages are sent in the correct order and progress never goes down.

PR #458, opened 2019-11-01
Feature delete mapping

gusmith:
The goal of this PR is to remove the `mapping` output type from the entity-service. The `mapping` output type is indeed redundant with the `groups` output type for 2 parties, however, the output representation is slightly different.

Few notes:
 - the documentation is horrible to update as there are a lot of things out of date.
 - the tutorial python notebooks are not manageable from here: we are mainly testing clkhash
 - most of the docs requires to have the entity-service already available (they point to `testing.es.data61.xyz`) which makes them impossible to keep up to date in the same PR as the one modifying the service

### DIFF
--- a/backend/entityservice/api_def/swagger.yaml
+++ b/backend/entityservice/api_def/swagger.yaml
@@ -40,7 +40,7 @@ info:
     The personally identifiable information used for linking is first locally transformed
     to anonymous linking codes called [Cryptographic Longterm Keys](concepts.html#cryptographic-longterm-key)
     (CLKs). Each party then uploads its CLKs to the service for matching.
-    The service supports four different [types of outputs](concepts.html#result-types)
+    The service supports three different [types of outputs](concepts.html#result-types)
     of matching results, varying in privacy properties.
     For the CLKs to be comparable, the parties need to agree on and follow a particular
     [linkage schema](concepts.html#schema) and agree on a shared secret before generating
@@ -68,7 +68,7 @@ info:
 
     ## Matching Protocols
 
-    The Entity Service supports four different **result types** with varying privacy properties which define the produced result,
+    The Entity Service supports three different **result types** with varying privacy properties which define the produced result,
     and who may see the which part of the output.
 
     See the documentation section on [output types](./concepts.html#result-types).
--- a/backend/entityservice/api_def/swagger.yaml
+++ b/backend/entityservice/api_def/swagger.yaml
@@ -75,13 +75,12 @@ info:
 
     The options are:
 
-      * `"mapping"` - Creates a lookup table of the form `indexA = indexB`.
       * `"permutations"` - Creates random permutations and a mask.
       * `"similarity_scores"` - Outputs a list of similarity scores of `[indexA, indexB, score]`, where `score`
         represents the likelihood that `indexA = indexB`.
       * `"groups"` - Outputs a list of groups of records, where each group represents one entity.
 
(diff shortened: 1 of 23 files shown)


backend/entityservice/tests/test_project_run_status.py
```diff
@@ -3,12 +3,17 @@
     create_project_no_data, ensure_run_progressing
 
 
-def test_run_status_with_clks_2p(requests, mapping_project):
-    size = mapping_project['size']
-    ensure_run_progressing(requests, mapping_project, size)
+def test_permutations_run_status_with_clks_2p(requests, permutations_project):
```
gusmith: I may not have understaood the reasons of these tests: it was done only on mapping but not on permutations and similarity scores... so I removed the mapping but add the two others.

backend/entityservice/tests/util.py
```diff
@@ -263,7 +263,7 @@ def wait_approx_run_time(size, assumed_rate=1_000_000):
     """
     number_comparisons = sum(
         size0 * size1 for size0, size1 in itertools.combinations(size, 2))
-    time.sleep(5)
+    time.sleep(1)
```
gusmith: Relates to issue #457

docs/development.rst
```diff
@@ -64,31 +65,6 @@ The run info ``HASH`` stores:
   ``backend/entityservice/cache/active_runs.py`` for implementation.
 
 
-Continuous Integration Testing
```
gusmith: I removed it because we should have before. And I added `devops` in the links from this page.

docs/future.rst
```diff
@@ -18,7 +18,7 @@ Road map for the entity service
 - optimise anonlink memory management and C++ code
 
 Bigger Projects
-- consider more than 2 organizations participating in one mapping
```
gusmith: I guess that's what the groups are doing.

backend/entityservice/tasks/permutation.py
```diff
@@ -113,8 +94,10 @@ def permute_mapping_data(project_id, run_id, len_filters1, len_filters2, parent_
         random.shuffle(remaining_new_indexes)
         log.info("Assigning random indexes for {} matched entities".format(number_in_common))
 
-        for mapping_number, a_index in enumerate(mapping):
-            b_index = mapping[a_index]
+        for mapping_number, group in enumerate(groups):
```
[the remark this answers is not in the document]
gusmith: Nicely spotted. :+1:

backend/entityservice/api_def/swagger.yaml
```diff
@@ -75,13 +75,12 @@ info:
 
     The options are:
 
-      * `"mapping"` - Creates a lookup table of the form `indexA = indexB`.
       * `"permutations"` - Creates random permutations and a mask.
       * `"similarity_scores"` - Outputs a list of similarity scores of `[indexA, indexB, score]`, where `score`
         represents the likelihood that `indexA = indexB`.
       * `"groups"` - Outputs a list of groups of records, where each group represents one entity.
 
-    Only `"groups"` supports multi-party linkage. `"mapping"`, `"permutations"`, and `"similarity_scores"` only support linkage
```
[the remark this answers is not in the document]
gusmith: :+1:

backend/entityservice/tests/util.py
gusmith: As I wasn't sure why it has been introduced, I reduced it to 1 second and opened an issue. I'll check what happens without this wait (f7601d179c07156d379af9a2dcd2081e3b8c2448), if the tests fail, I'll revert it, otherwise keep it.

PR #461, opened 2019-11-05
release 1.13.0-alpha

wilko77:
I jump from .12 to .13 because of the breaking changes.


docs/changelog.rst
```diff
@@ -4,7 +4,7 @@
 Changelog
 =========
 
-Next Version
```
gusmith: I think it's good to keep the `next version` section, pointing out that we should always update the changelog.

wilko77: I am following the process in the `realease.md` file.
The 'Next Version' section will be re-introduced in the second to last step.

PR #464, opened 2019-11-13
Feature similarity groups

gusmith:
This is a first step to have the similarity score output for multi-party: this PR modifies the output from 
```
[
  [row_index_0, row_index_1, score],
  ...
]
```
to
```
[
  {'group': [[party_id_0, row_index_0], [party_id_1, row_index_1]], 'sim': score},
  ...
]
```
which follows the `groups` output format of the candidate pairs (i.e. a record is represented by two indices `[party_id, row_index]`.


backend/entityservice/api_def/swagger.yaml
```diff
@@ -509,25 +509,26 @@ paths:
 
         ### result_type = "similarity_scores"
 
-        The list of the indices of potential matches and their similarity score
+        The list of the candidate of potential matches and their similarity score
```
wilko77: How about: "The list of the pairwise similarity scores where the score is greater than the similarity threshold."
(when we create a run we talk about 'similarity threshold', not about 'mapping threshold'.)
I don't think we should talk about potential matches here, as we return everything.

backend/entityservice/api_def/swagger.yaml
```diff
@@ -509,25 +509,26 @@ paths:
...
         where the similarity score is greater than the mapping threshold.
         Data is returned as `json` object e.g.,
 
             {
                 "similarity_scores":
                     [
-                        [5, 27, 1.0],
-                        [14, 10, 1.0]
+                        {'group': [[0, 5], [1, 27]], 'sim': 1.0},
+                        {'group': [[1, 10], [0, 14]], 'sim': 1.0}
```
wilko77: This format has a lot of overhead, as it is writing 'group' and 'sim' for every pair.
How about 
```
{
  "similarity_scores":
  [
    [[0, 5], [1, 27], 1.0], 
    [[1, 10], [0, 14], 1.0]
  ]
}

also, I don't think 'group' is a good choice, as when we extend to the multiparty case, this gets really confusing.
The groups result shows actual groups of 2 or more entities, whereas this will only ever show pairwise results.

Comment (gusmith) [reply backend/entityservice/api_def/swagger.yaml]: I didn't realise it was valid JSON to have different value types in a single array, but it reduces the size of the output drastically, and simplify the discussion about the keys to use: none :)

File: backend/entityservice/api_def/swagger.yaml (hunk 1 above)
Comment (gusmith) [reply backend/entityservice/api_def/swagger.yaml]: :+1:

File: docs/changelog.rst [hunk 3]
```diff
@@ -8,14 +8,14 @@ Next Version
...
-- modify `similiraty_score` output to follow the group format, which will simplify extending this output type to more parties (#464)
+- modify ``similiraty_score`` output to follow the group format, which will simplify extending this output type to more parties (#464)
 
 Breaking Change
 ~~~~~~~~~~~~~~~
 
 - the ``dataproviders`` table `uploaded` field has been modified from a BOOL to an ENUM type (#463)
-- the `similiraty_score` output type has been modified, it now returns a JSON array of JSON objects, where such an object
-  looks like `{'group': [[party_id_0, row_index_0], [party_id_1, row_index_1]], 'sim': score}`.
+- the ``similiraty_score`` output type has been modified, it now returns a JSON array of JSON objects, where such an object
```
wilko77: same typo :)

backend/entityservice/api_def/swagger.yaml
```diff
@@ -509,26 +509,27 @@ paths:
...
+        refers to the record at the index `row_index_0` from the dataset `party_id_0`, similarly for
+        `[party_id_1, row_index_1]`, and `score` is the similarity score representing the likelihood
+        that this pair or records is a match.
 
         `ds_index_0`, `rec_index_0, `ds_index_1` and `rec_index_1` start from 0, and `party_id_0 != party_id_1` but
-        not necessarilly ordered.
+        are not necessarilly ordered.
 
         The value of `score` is between 0.0 and 1.0, where 0.0 corresponds to no match
         and 1.0 corresponds to total match.
```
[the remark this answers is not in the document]
gusmith: :+1:

docs/changelog.rst
gusmith: :+1:

gusmith:
The tutorials are currently faliing because of the newest release of clkhash which has some breaking changes.
The PR #465 is fixing them, but in progress.

gusmith:
@wilko77 Since your review, I simply merged dev into this branch (resolving some conflicts in a tutorial), and added a note about the tutorials usage, mainly because the ones from this branch are NOT working with the currently deployed service (because of some breaking change in the code which has not been deployed).

PR #466, opened 2019-11-19
Remove redundant tests

### LINKED ISSUE #457: tests `test_project_run_status` are currently useless (or redundant)

gusmith:
This PR will close #457 
Currently, the entity-service is trying to do unit tests in an end to end way. For example, the removed tests were about testing the run endpoint by creating full runs.
These full runs are already created in a number of other tests (e.g. the ones where we check the validity of the output of a run).
So in this PR, I removed some tests about intermediary results such as checking that a run is progressing normally and testing the run endpoint.

However, this is slightly against the tests visions where each test describe what it is "unit" testing. In the approach I started, we run the whole end to end pipeline and on the way test each part. 

But from 327 tests running in 7m 40s on my machine with a simple local deployment on the service, the PR reduced the number of tests to 283 ran in 6m 35s with the same deployment.

### DIFF
--- a/backend/entityservice/tests/util.py
+++ b/backend/entityservice/tests/util.py
@@ -234,7 +234,7 @@ def wait_for_run(requests, project, run_id, ok_statuses, result_token=None, time
     Poll project/run_id until its status is one of the ok_statuses. Raise a
     TimeoutError if we've waited more than timeout seconds.
     It also checks that the status are progressing normally, using the checks implemented in
-    `has_not_progressed_invalidly`.
+    `has_progressed_validly`.
     """
     start_time = time.time()
     old_status = None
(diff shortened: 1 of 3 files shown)


backend/entityservice/tests/util.py
```diff
@@ -362,7 +350,7 @@ def from_string(state):
             return State.completed
 
 
-def has_not_progressed_invalidly(status_old, status_new):
+def has_progressed_validly(status_old, status_new):
```
wilko77: I do not not approve of this change ;)

PR #473, opened 2020-02-02
Updates production deployment documentation.

### LINKED ISSUE #471: Update production deployment docs
Verify links and modify for helm v3.

hardbyte:
This started as just updating a dead hyperlink but has ended up being misc updates to the Kubernetes deployment docs.

Closes #471

### DIFF
--- a/deployment/jobs/integration-test-job.yaml
+++ b/deployment/jobs/integration-test-job.yaml
@@ -15,11 +15,11 @@ spec:
       restartPolicy: Never
       containers:
       - name: entitytester
-        image: quay.io/n1analytics/entity-app
+        image: data61/anonlink-app:v1.12.0
         imagePullPolicy: Always
         env:
           - name: ENTITY_SERVICE_URL
-            value: https://beta.anonlink.data61.xyz/api/v1
+            value: https://anonlink.easd.data61.xyz/api/v1
           - name: LOGGING_LEVEL
             value: "INFO"
         command:
--- a/docs/production-deployment.rst
+++ b/docs/production-deployment.rst
@@ -1,13 +1,15 @@
 Production deployment
 =====================
 
-Production deployment assumes a multi node `Kubernetes <https://kubernetes.io/docs/home/>`__
+Production deployment assumes a `Kubernetes <https://kubernetes.io/docs/home/>`__
 cluster.
 
-The entity service has been deployed to kubernetes clusters on Azure, GCE, minikube and
+The entity service has been deployed to kubernetes clusters on Azure, GCE, minikube, and
 AWS. The system has been designed to scale across multiple nodes and handle node
 failure without data loss.
 
+Overview
+--------
 
 .. figure:: _static/deployment.png
    :alt: Entity Service Kubernetes Deployment
@@ -15,8 +17,8 @@ failure without data loss.
 
 At a high level the main custom components are:
 
-- **ES App** - a gunicorn/flask backend web service hosts the REST api
-- **Entity Match Worker** instances - uses celery for task scheduling
+- **REST API Server** - a gunicorn/flask backend web service hosting the REST api.
+- **PPRL Worker** instances - using celery for task scheduling.
 
 The components that are used in support are:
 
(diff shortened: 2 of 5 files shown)


docs/production-deployment.rst
```diff
@@ -151,34 +148,54 @@ Then, it advised to use the `--dry-run --debug` options before deploying with `h
 Installation
 ~~~~~~~~~~~~
 
-To install the whole system execute::
+To install the whole system assuming you have a configuration file ``my-deployment.yaml`` in the current
+directory::
 
-    cd deployment
-    helm install entityservice --name="anonlink" --values ``my-deployment.yaml``
+    $ helm upgrade --install anonlink data61/entity-service -f anonlink.yaml
```
wilko77: shouldn't that be `my-deployment.yaml` instead of `anonlink.yaml`? Or is there some strange magic involved?

docs/production-deployment.rst
```diff
@@ -191,14 +208,14 @@ to 20::
...
-that in version control.
+However, note you may wish to instead keep all configurable values in a ``yaml`` file and track
+the changes in version control.
 
 Minimal Deployment
 ------------------
 
-To run with minikube for local testing we have provided a ``minimal.yaml`` file that will
-set very small resource limits. Install the minimal system with::
+To run with minikube for local testing we have provided a ``minimal.yaml`` configuration file that will
```
wilko77: where is that `minimal.yaml` file? I couldn't find it in the deployment folder.

docs/production-deployment.rst
```diff
@@ -191,14 +208,14 @@ to 20::
     helm upgrade anonlink entity-service --namespace=testing --set workers.replicas="20"
```
wilko77: a few lines before you used `data61/entity-service`. Now only `entity-service`. Is that deliberate?

deployment/jobs/integration-test-job.yaml
```diff
@@ -15,11 +15,11 @@ spec:
       restartPolicy: Never
       containers:
       - name: entitytester
-        image: quay.io/n1analytics/entity-app
+        image: data61/anonlink-app:v1.12.0
         imagePullPolicy: Always
         env:
           - name: ENTITY_SERVICE_URL
-            value: https://beta.anonlink.data61.xyz/api/v1
+            value: https://anonlink.easd.data61.xyz/api/v1
```
wilko77: why do we define the server differently to the benchmark yaml? Is there a reason for this inconsistency?
Wouldn't it be nicer if they both accepted the same name/value pair?

deployment/jobs/integration-test-job.yaml
```diff
@@ -15,11 +15,11 @@ spec:
       restartPolicy: Never
       containers:
       - name: entitytester
-        image: quay.io/n1analytics/entity-app
+        image: data61/anonlink-app:v1.12.0
```
wilko77: wouldn't it be better to link to `latest` or `stable` or something similar. We don't want to update the deployment files with every release.

docs/production-deployment.rst
hardbyte: No magic. Good catch

docs/production-deployment.rst
hardbyte: Hmm, I would have sworn there was one. I'll remove the reference as it was so minimal it couldn't do much except deploy the service

docs/production-deployment.rst
hardbyte: Will change - the issue is because previously all the docs assumed you were deploying from source code.

deployment/jobs/integration-test-job.yaml
hardbyte: Sure would be -> #475

deployment/jobs/integration-test-job.yaml
hardbyte: Sure, happy to use `latest`. It is mostly there as an example, and pinning exact versions of what runs is considered best practice.

hardbyte: I'll remove the log level setting from here too - it is just for the logging level during tests which pytest overrides anyway.

hardbyte:
Thanks for the review @wilko77 - have addressed the inconsistencies in the documentation and opened an issue for the env var.

PR #476, opened 2020-02-03
Update ci cluster

hardbyte:
Refactors most of the Kubernetes related CI testing to use azure's service connections and to migrate to our newer cluster.


azure-pipelines.yml
```diff
@@ -128,167 +131,183 @@ stages:
...
+      - task: Kubernetes@1
+        displayName: 'Wait for integration test pod to be ready'
+        name: waitpods
+        inputs:
+          connectionType: 'Kubernetes Service Connection'
+          kubernetesServiceEndpoint: $(CLUSTER)
+          command: 'wait'
+          arguments: '--namespace $(NAMESPACE) -ldeployment=$(DEPLOYMENT) --timeout=30s --for=condition=Ready pods'
+
+      - task: Kubernetes@1
```
JoyceXinyueWang: Just wondering what is the difference between Kubernetes@0 and Kubternetes@1?

hardbyte:
> These is the most complicated azure pipeline yaml file I ever seen
 :-( Perhaps it needs to be split further into templates. I suggest merging so we can rebase the other work and keep developing though.

The deployment warning is fine - I've seen that when the redis self test pod doesn't work.

azure-pipelines.yml
hardbyte: Versions of the task - if there is a `Kubernetes@0` I should update it

hardbyte: Looks like they are fine. If you're interested in the docs have a look [here](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/deploy/kubernetes?view=azure-devops)

PR #479, opened 2020-02-05
Feature openapi3

wilko77:
ok, this all started with me wanting to extend the clk upload endpoint to alternatively accept clk'n'blocks. 

I think that the `oneOf` operator would be nice and elegant. However, this is only supported from openapi 3 onwards. 

In version 3, the clks endpoint would then look like this:
```
'/projects/{project_id}/clks':
    post:
      operationId: entityservice.views.project.project_clks_post
      summary: Upload encoded PII data to a linkage project.

      description: ...

      parameters:
        - $ref: '#/components/parameters/project_id'
        - $ref: '#/components/parameters/token'
      requestBody:
        description: the clks as json
        required: true
        content:
          application/json:
            schema:
              oneOf:
                - $ref: '#/components/schemas/CLKUpload'
                - $ref: '#/components/schemas/CLKnBlockUpload'
          application/octet-stream:
            schema:
              type: string
              format: binary
```
Unfortunately, _connexion_ is not helping.
Currently, if an endpoint is defined to accept more than one kind of content types, connexion 
ignores this information and assumes that they are all `application/json`. See for instance [here](https://github.com/zalando/connexion/issues/860).
There is some [work in progress](https://github.com/zalando/connexion/pull/760) that started in 2018 to address this.

So for now, I addressed this by creating a separate endpoint (binaryclks) to accept the binary format.

Another annoyance: I added the swaggerUI, thought that would be handy to play with, and read about the api. Unfortunately, connexion does some [magic](https://github.com/zalando/connexion/issues/1139) to the `servers` definition in the api spec such that (in our case) they disappear.

Can you have a look at this so far and tell me what you think?


wilko77:
Another connexion weirdness:

I defined the CLKnBlocks upload like this:
```
CLKnBlockUpload:
      description: Array of this party's Bloom Filters including blocking information
      type: object
      required: [clknblocks]
      properties:
        clknblocks:
          type: array
          items:
            type: array
            items:
              allOf:
                - type: string
                  format: byte
                  description: Base64 encoded CLK data
                - type: array
                  items:
                    oneOf:
                      - type: string
                      - type: integer
```
connexion is not impressed though:
```
'clknblocks':[['IugCCIixXAEECgS4LCUUREgswuUBaIAwiqJIQAAZqsCAjLYBCOkBABACgOWPpngwhloFgwDLEj1RIKMQAC1ALWIU5QQSqAQiQKBvZAIOKDFEK4igATRIAJcAGiRYGglYEBA1AjhAHFATu1VBRV49hEQIJDxVEisWDyQ0DJDaXIg=', ['000011010110100111100', '111100000101001010100', '200011010001000001001', '310000001000000000001', '400011001100111100010', '501000000100001101110', '610000000111001010010', '700001101011000000000', '800000000100001001010', '900000000101000101000', '1001000101000110000000', '1100000010110100000001', '1210011010000100010010', '1300100011110010100100', '1401000100100000000010', '1510010101100000011010', '1601000011010010010010', '1700110000000010001000', '1800000000000000000101', '1910000010000001110000', '2000000010010010011001', '2100000010010000000011', '2201100101110110110000', '2310010100000110001000', '2411001000000110000000', '2500100010000001000000', '2601110010010000101000', '2710001010100000100101', '2811000000111000000010', '2900010100000100100000']]]} is not valid under any of the given schemas
```

If I simplify the definition to this:
```
CLKnBlockUpload:
      description: Array of this party's Bloom Filters including blocking information
      type: object
      required: [clknblocks]
      properties:
        clknblocks:
          type: array
          items:
            type: array
            items:
              anyOf:
                - type: string
                  format: byte
                  description: Base64 encoded CLK data
                - type: string
                - type: integer
```
it works. However, this definition is not as clean, as it does not require to have either clk or block info.

wilko77:
OK, I officially declare this branch ready for review.

Main work:
- I separated binary from json upload, because connexion does not support multiple content types This has the advantage, that for the binary upload endpoint, we can now define and enforce the extra headers.
- the json upload endpoint expects either a `CLKUpload` or `CLKnBlocksUpload` request body. As described in the comment above, I am not totally happy with the CLKnBlocksUpload definition. If someone can tame connexion to accept a better format, I'm all up for it.

backend/requirements.txt
```diff
@@ -3,7 +3,7 @@ bitmath==1.3.1.2
 celery==4.4.0
 clkhash==0.15.1
 colorama==0.4.1 # required for structlog
-connexion[swagger-ui]==2.6
+connexion==2.6.0
```
hardbyte: I thought we saw warnings logged if we didn't include `swagger-ui`?

backend/entityservice/api_def/openapi.yaml
```diff
@@ -0,0 +1,1078 @@
...
+info:
+  version: '1.3'
+  title: Entity Matching API
+  x-logo:
+    url: http://clkhash.readthedocs.io/en/latest/_static/logo.svg
+  contact:
+    name: 'Confidential Computing, Data61 | CSIRO'
+    email: confidential-computing@csiro.au
+  description: >-
+    Allows two organisations to carry out private record linkage -
```
hardbyte:
```suggestion
    Allows multiple organisations to carry out private record linkage -
```

backend/entityservice/api_def/openapi.yaml
```diff
@@ -0,0 +1,1078 @@
...
+      summary: Upload encoded PII data to a linkage project.
+      tags:
+        - Project
+      description: |
+        Called by each of the data providers with their calculated `CLK` vectors.
+        The project must have been created, and the caller must have both the
+        `project_id` and a valid `upload_token` in order to contribute data.
+
+        The data uploaded must be of one of the following formats.
+        - CLKs only upload: An array of base64 encoded CLKs (./concepts.html#cryptographic-longterm-keys), one per
```
hardbyte: Use markdown for the link

backend/entityservice/api_def/openapi.yaml
```diff
@@ -0,0 +1,1078 @@
...
+          $ref: '#/components/responses/RateLimited'
+
+  '/projects/{project_id}/binaryclks':
+    post:
+      operationId: entityservice.views.project.project_binaryclks_post
+      summary: Upload binary encoded PII data to a linkage project.
+      tags:
+        - Project
+      description: |
+        An experimental api has been added for uploading CLKs as a binary file. This is to allow for
```
hardbyte:
```suggestion
        An experimental api for uploading CLKs as a binary file. This is to allow for
```

backend/entityservice/api_def/openapi.yaml
```diff
@@ -0,0 +1,1078 @@
...
+      properties:
+        clks:
+          type: array
+          items:
+            type: string
+            format: byte
+            description: Base64 encoded CLK data
+
+    CLKnBlockUpload:
+      description: Array of this party's Bloom Filters including blocking information
```
hardbyte: This is an object containing a single element which is an array. Do we need/want the extra layer just so you can say `clknblocks`?

backend/requirements.txt
wilko77: two options:
1. disable swagger-ui. 
```
con_app.add_api(pathlib.Path("openapi.yaml"),
                base_path='/',
                options={'swagger_ui': False},
                strict_validation=config.CONNEXION_STRICT_VALIDATION,
                validate_responses=config.CONNEXION_RESPONSE_VALIDATION)
```
2. use swagger_ui. For this, we have to solve two problems. First the default url to the spec is not working in our case.
```
con_app.add_api(pathlib.Path("openapi.yaml"),
                base_path='/',
                options={'swagger_ui_config': {'url': '/api/v1/openapi.json'}},
                strict_validation=config.CONNEXION_STRICT_VALIDATION,
                validate_responses=config.CONNEXION_RESPONSE_VALIDATION)
```
Yay. Secondly, connexion removes the server description/url. (See above) Therefore you cannot actually use the UI to hit the endpoints.

Because 2 doesn't work, I favor 1.

backend/entityservice/api_def/openapi.yaml
wilko77: With the clks only, we upload a json of the form {"clks": [...]}.
Therefore I thought that for better separation, we will give the new format a new key.
That simplifies figuring out the format on the receiving end.
I also like to have the clks under a key and not top level, such that we can easily add meta info without breaking anything.

backend/requirements.txt
hardbyte: Also in favor of option 1 👍

backend/entityservice/api_def/openapi.yaml
hardbyte: It would still be nice for the description to be accurate - i.e. `object` rather than `Array`

PR #490, opened 2020-02-07
Fix dependencies and migrate benchmark auth

hardbyte:
This PR makes a few small changes to the benchmark container/script:
- If an object store bucket name is given we will try write results to it - even if no ACCESS_KEY is given. this is to allow deployments on e.g. AWS where Ec2 nodes can be granted permission to do things.
- Updates the benchmark script to use the new clkhash `rest_client` api which.
- Patch version bump to the benchmark container

This depandabot flurry of dependency updates somehow managed to update Flask to a version that isn't compatible with our code. The integration tests passed on the PR, but not once in develop 👎 
So I've also upped the version of connexion just to get the CI to pass, in theory this would be done in the PR #479 that Wilko is working on.


backend/requirements.txt
```diff
@@ -1,9 +1,9 @@
 anonlink==0.12.5
 bitmath==1.3.1.2
 celery==4.4.0
-clkhash==0.15.0
+clkhash==0.15.1
 colorama==0.4.1 # required for structlog
-connexion==1.4
+connexion[swagger-ui]==2.6
```
JoyceXinyueWang: This looks a bit weird..

benchmarking/benchmark.py
```diff
@@ -24,7 +24,7 @@
 import time
 import os
 
-from clkhash import rest_client
+from anonlinkclient.rest_client import RestClient
```
JoyceXinyueWang: 💯

benchmarking/requirements.txt
```diff
@@ -1,6 +1,6 @@
+anonlink-client==0.0.1
```
JoyceXinyueWang: This will be 0.1.0 pretty soon

benchmarking/benchmark.py
```diff
@@ -232,11 +234,11 @@ def compose_result(status, tt, experiment, sizes, threshold):
     return result
 
 
-def delete_resources(config, credentials, run):
+def delete_resources(credentials, run):
```
JoyceXinyueWang: it would be great to see what key-value pairs might be in `config` in doc

benchmarking/benchmark.py
```diff
@@ -232,11 +234,11 @@ def compose_result(status, tt, experiment, sizes, threshold):
...
     try:
         if run is not None and 'run_id' in run:
-            rest_client.run_delete(config['server'], credentials['project_id'], run['run_id'], credentials['result_token'])
-        rest_client.project_delete(config['server'], credentials['project_id'], credentials['result_token'])
+            rest_client.run_delete(credentials['project_id'], run['run_id'], credentials['result_token'])
+        rest_client.project_delete(credentials['project_id'], credentials['result_token'])
```
JoyceXinyueWang: Is it true that currently there is only server in `config`?

backend/requirements.txt
hardbyte: I'll say!

benchmarking/requirements.txt
hardbyte: Sweet. Our depandabot will pester us when it is released :-)

benchmarking/benchmark.py
hardbyte: No, there is lots of stuff in `config` all set around line 84. the change is because the new `rest_client` already knows the server address

benchmarking/benchmark.py
```diff
@@ -314,39 +315,39 @@ def run_experiments(config):
...
     credentials = {}
     run = {}
     logger.info("Starting time: {}".format(time.asctime()))
     nb_parties = len(sizes)
     try:
-        credentials = rest_client.project_create(server, config['schema'], 'groups',
+        credentials = rest_client.project_create(config['schema'], 'groups',
                                                  "benchy_{}".format(experiment), parties=nb_parties)
         # upload clks
         upload_binary_clks(config, sizes, credentials)
```
wilko77: ideally, that should be done by the rest_client as well.

benchmarking/benchmark.py
```diff
@@ -314,39 +315,39 @@ def run_experiments(config):
...
                                      threshold,
                                      "{}_{}".format(experiment, threshold))
         # wait for result
         run_id = run['run_id']
         result['run_id'] = run_id
         logger.info(f'waiting for run {run_id} from the project {project_id} to finish')
-        status = rest_client.wait_for_run(server, project_id, run_id,
-                                          credentials['result_token'], timeout=config['timeout'])
+        status = rest_client.wait_for_run(project_id, run_id,
+                                          credentials['result_token'], timeout=config['timeout'], update_period=5)
```
wilko77: will that lead to a new line in the logs every 5 seconds? what about those huge benchmarks? How big's the log file going to be?

benchmarking/requirements.txt
```diff
@@ -1,6 +1,6 @@
...
 arrow
 boto3
-clkhash==0.14.0
 jsonschema
 pandas
 requests
```
wilko77: shoule we pin all of those as well?

deployment/jobs/benchmark/timing-test-job.yaml
```diff
@@ -35,16 +35,18 @@ spec:
...
-            valueFrom:
-              secretKeyRef:
-                name: anonlink-benchmark-aws-credentials
-                key: OBJECT_STORE_ACCESS_KEY
-          - name: OBJECT_STORE_SECRET_KEY
-            valueFrom:
-              secretKeyRef:
-                name: anonlink-benchmark-aws-credentials
-                key: OBJECT_STORE_SECRET_KEY
+          - name: OBJECT_STORE_BUCKET
```
wilko77: don't you also have to define `OBJECT_STORE_SERVER`?

hardbyte: No we default to assuming S3 is used

benchmarking/benchmark.py
hardbyte: Not on the benchmark container side - we just wait until the whole thing is done (perhaps we should be outputting progress but that is another question).

On the server side we would log the request in both nginx and the flask app.

benchmarking/benchmark.py
hardbyte: Agree => https://github.com/data61/anonlink-client/issues/23

PR #500, opened 2020-02-12
Revive jaeger

hardbyte:
Brings Jaeger/opentracing support back to life.

Refactors some of the logging config to be reused in the tracing config. I just pulled out the yaml config file loader and added tests.

Restructures the config via k8s to be a bit simpler - now all monitoring related config files are mounted in `/var/config`, the filenames are still passed in via ENV vars.

The default `values.yaml` file for our helm chart now includes annotations that will enable opentracing via a jaeger sidecar - if the cluster supports it. It has no effect on clusters that don't have jaeger enabled so I don't think it is a big deal to enable by default.

I uncommented the logging config in the values.yaml file so it is actually tested/used. It needs improvement.

### DIFF
--- a/backend/entityservice/tests/test_utils.py
+++ b/backend/entityservice/tests/test_utils.py
@@ -1,0 +1,64 @@
+import textwrap
+
+import pytest
+
+from entityservice.errors import InvalidConfiguration
+from entityservice.utils import load_yaml_config
+from entityservice.tests.util import generate_bytes, temp_file_containing
+
+
+class TestYamlLoader:
+
+    def test_empty(self):
+        with temp_file_containing(b'') as fp:
+            filename = fp.name
+            assert None == load_yaml_config(filename)
+
+    def test_list(self):
+        with temp_file_containing(b'[1,2,3]') as fp:
+            filename = fp.name
+            assert [1,2,3] == load_yaml_config(filename)
+
+    def test_missing_file(self):
+        filename = 'unlikely a valid file'
+        with pytest.raises(InvalidConfiguration):
+            load_yaml_config(filename)
+
+    def test_random_bytes(self):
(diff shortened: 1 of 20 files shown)


backend/entityservice/tests/test_utils.py
```diff
@@ -0,0 +1,64 @@
...
+            filename = fp.name
+            assert [1,2,3] == load_yaml_config(filename)
+
+    def test_missing_file(self):
+        filename = 'unlikely a valid file'
+        with pytest.raises(InvalidConfiguration):
+            load_yaml_config(filename)
+
+    def test_random_bytes(self):
+        with temp_file_containing(generate_bytes(128)) as fp:
```
wilko77: there is a non-zero chance that those random bytes resemble valid yaml. Why not just feeding a known invalid config to the test?

backend/entityservice/tests/util.py
```diff
@@ -14,6 +15,15 @@
 from entityservice.tests.config import url
 
 
+@contextmanager
+def temp_file_containing(data):
+    # Code to acquire resource, e.g.:
```
wilko77: that is not a very useful comment

backend/entityservice/utils.py
```diff
@@ -8,15 +8,36 @@
...
+
+    :param filename: a Path or String object
+    :raises InvalidConfiguration if the file isn't found or the yaml isn't valid.
+    :return: Python representation of yaml file's contents (usually Dict), or None if empty.
+    """
+    try:
+        with open(filename, 'rt') as f:
+            return yaml.safe_load(f)
+    except UnicodeDecodeError as e:
+        raise InvalidConfiguration("YAML file appears corrupt") from e
```
wilko77: it would be nice to include the filename in the error message.

PR #503, opened 2020-02-17
Extend upload endpoint to accept both types - clks and clknblocks

### LINKED ISSUE #501: Extend upload endpoint to accept all formats specified in openAPI
That's the file views/project.py.
Method project_clks_post needs to support both upload types.
The plan is to give the new format the same treatment as the old one.
That is:
- take the data as-is and save it in a file.
- call the handle_raw_upload task to deal with it. (**a separate issue**)

Additionally, we would also have to check that the uploaded data is what we expect. Either the json has a 'clks' element, or a 'clknblocks' one. That has to match the 'use_blocking' flag in the project info in the database.

JoyceXinyueWang:
By introducting blocking in `anonlink-client`, we also want to update `entity-service` to support blocking.

In this pull request, we extended the upload endpoint i.e. `upload_json_clk_data` to accept both uploading types:
- json with `clks` element
- json with `clknblocks` element
with a focus on error handling. 

Note that we also changed the way that writes to MINIO from `put_object` to `fput_object` to avoid large in memory bytes conversion. 

Currently it will fail CI since it is only the start of this update to support blocking and more modifications are required to make the whole entity service work with blocking.

Close #501


backend/entityservice/views/project.py
```diff
@@ -347,39 +353,65 @@ def upload_clk_data_binary(project_id, dp_id, raw_stream, count, size=128):
...
+
+    # write clk_json into a temp file
+    _, tmp_filename = tempfile.mkstemp()
 
-        num_bytes = len(data)
-        span.set_tag("num_bytes", num_bytes)
-        buffer = BytesIO(data)
+    logger.info('Writing uploaded {} JSON to {}'.format(element.upper(), tmp_filename))
+    with open(tmp_filename, 'w') as f:
+        json.dump(clk_json, f)
```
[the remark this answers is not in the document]
JoyceXinyueWang: Sure. Should we create a separate branch for blocking support development?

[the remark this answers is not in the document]
hardbyte: sounds good to me. Perhaps make an empty branch based off develop called `feature-blocking` and change the merge target for this PR?

backend/entityservice/views/project.py
```diff
@@ -217,7 +219,11 @@ def project_clks_post(project_id):
                 #       consume the stream first to validate it against the spec. Thus the backflip to fully reading the CLks as
                 #       json into memory. -> issue #184
 
-                receipt_token, raw_file = upload_json_clk_data(dp_id, get_json(), span)
+                # get flag use_blocking from table projects
+                with DBConn() as conn:
```
wilko77: every time you call `with DBConn() as conn:` it will request a database connection. This creates some overhead. Thus it is better to group database interactions. About 20 lines above, we already had a database connection to do some queries. You could use that one instead of requesting a new one here.

backend/entityservice/views/project.py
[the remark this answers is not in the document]
JoyceXinyueWang: Okay I will do that

backend/entityservice/views/auth_checks.py
```diff
@@ -59,6 +59,31 @@ def abort_if_invalid_results_token(resource_id, results_token):
         safe_fail_request(403, message=INVALID_ACCESS_MSG)
 
 
+def abort_if_inconsistent_upload(uses_blocking, clk_json):
+    # check if the following combinations are true
+    # - uses_blocking is False AND 'clks' element in upload JSON
+    # - uses_blocking if True AND 'clknblocks' element in upload JSON
+    # otherwise, return safe_fail_request
+    is_valid_clks = not uses_blocking and 'clks' in clk_json
+    is_valid_clknblocks = uses_blocking and 'clknblocks' in clk_json
```
hardbyte: Thanks I think this is much easier to understand

backend/entityservice/views/auth_checks.py
```diff
@@ -59,6 +59,31 @@ def abort_if_invalid_results_token(resource_id, results_token):
         safe_fail_request(403, message=INVALID_ACCESS_MSG)
 
 
+def abort_if_inconsistent_upload(uses_blocking, clk_json):
+    # check if the following combinations are true
+    # - uses_blocking is False AND 'clks' element in upload JSON
+    # - uses_blocking if True AND 'clknblocks' element in upload JSON
+    # otherwise, return safe_fail_request
```
hardbyte: Nitpick: This might as well be a docstring. (code comments like this don't come up IDE tooltips)

backend/entityservice/views/project.py
```diff
@@ -198,6 +200,8 @@ def project_clks_post(project_id):
         dp_id = db.get_dataprovider_id(conn, token)
         project_encoding_size = db.get_project_schema_encoding_size(conn, project_id)
         upload_state_updated = db.is_dataprovider_allowed_to_upload_and_lock(conn, dp_id)
+        # get flag use_blocking from table projects
+        uses_blocking = get_project_column(conn, project_id, 'uses_blocking')
```
hardbyte: 👍

backend/entityservice/views/project.py
```diff
@@ -216,8 +220,7 @@ def project_clks_post(project_id):
                 #       However, as connexion is very, very strict about input validation when it comes to json, it will always
                 #       consume the stream first to validate it against the spec. Thus the backflip to fully reading the CLks as
                 #       json into memory. -> issue #184
-
-                receipt_token, raw_file = upload_json_clk_data(dp_id, get_json(), span)
+                receipt_token, raw_file = upload_json_clk_data(dp_id, get_json(), span, uses_blocking)
```
hardbyte: Another nitpick. Utility arguments such as `span` should come after the useful ones.

backend/entityservice/views/project.py
```diff
@@ -347,39 +350,44 @@ def upload_clk_data_binary(project_id, dp_id, raw_stream, count, size=128):
     return receipt_token
 
 
-def upload_json_clk_data(dp_id, clk_json, parent_span):
+def upload_json_clk_data(dp_id, clk_json, parent_span, uses_blocking):
     """
     Convert user provided encodings from json array of base64 data into
     a newline separated file of base64 data.
```
hardbyte: I think this docstring is now out of date?

backend/entityservice/views/project.py
```diff
@@ -347,39 +350,44 @@ def upload_clk_data_binary(project_id, dp_id, raw_stream, count, size=128):
...
+    logger.info("Storing user {} supplied {} from json".format(dp_id, element))
 
     with opentracing.tracer.start_span('splitting-json-clks', child_of=parent_span) as span:
-        count = len(clk_json['clks'])
-        span.set_tag("clks", count)
-        data = b''.join(''.join(clk.split('\n')).encode() + b'\n' for clk in clk_json['clks'])
+        count = len(clk_json[element])
+        span.set_tag(element, count)
+
+    logger.info(f"Received {count} encodings.")
```
hardbyte: If the element type was `clksnblocks` will this statement be correct?

backend/entityservice/views/project.py
```diff
@@ -347,39 +350,44 @@ def upload_clk_data_binary(project_id, dp_id, raw_stream, count, size=128):
...
+    # write clk_json into a temp file
+    tmp = tempfile.NamedTemporaryFile(mode='w', delete=False)
 
-        num_bytes = len(data)
-        span.set_tag("num_bytes", num_bytes)
-        buffer = BytesIO(data)
+    logger.info('Writing uploaded {} JSON to {}'.format(element.upper(), tmp.name))
+    json.dump(clk_json, tmp)
 
-    logger.info(f"Received {count} encodings. Uploading {fmt_bytes(num_bytes)} to object store")
```
hardbyte: This now looks like a debug log statement - the `tmp.name` won't actually be around after the tmp file goes out of scope.

Could be worth logging the object store `filename` at info level in here though.

backend/entityservice/views/project.py
wilko77: yes

hardbyte:
Closing after rebasing into `feature-blocking`

PR #505, opened 2020-02-17
Feature extend binary format

wilko77:
This PR changes the binary format for storing clks in the object store.

Previously, we implicitly inferred the entity ID from the index of the CLK in the list. That was OK, as we had the CLKs delivered in order. With the blocking however, we don't have this order any more, and thus, we have to keep track which CLK belongs to which entity ID in order to produce an interpret-able result...

I chose to encode the entity ID as an unsigned int. I know, there are more people on the earth than that, but hey, storage costs money...


backend/entityservice/database/selections.py
```diff
@@ -260,32 +260,36 @@ def get_dataprovider_id(db, update_token):
...
+        """.format(', '.join(columns))
     result = query_db(db, sql_query, [dp_id], one=True)
     if result is None:
         raise DataProviderDeleted(dp_id)
     else:
-        return result[column]
+        if len(columns) > 1:
+            return [result[column] for column in columns]
+        else:
+            return result[columns[0]]
```
hardbyte: Not 100% enjoying returning a different type for convenience when there is only one column. I get it & I've done it I just don't like it.

backend/entityservice/serialization.py
```diff
@@ -41,11 +41,12 @@ def binary_format(encoding_size):
 
     The binary format string can be understood as:
     - "!" Use network byte order (big-endian).
+    - "I" store the entity ID as an unsigned int
```
hardbyte: I guess we aren't going to be handling 4 billion entities in a single run anytime soon.

backend/entityservice/serialization.py
```diff
@@ -54,31 +55,41 @@ def binary_pack_filters(filters, encoding_size):
...
+            - second element is 'encoding_size' number of bytes as produced by deserialize_bytes.
 
     :return:
         An iterable of bytes.
     """
     bit_packing_struct = binary_format(encoding_size)
 
     for hash_bytes in filters:
-        yield bit_packing_struct.pack(hash_bytes)
+        yield bit_packing_struct.pack(*hash_bytes)
```
hardbyte: ⭐️

backend/entityservice/serialization.py
```diff
@@ -41,11 +41,12 @@ def binary_format(encoding_size):
 
     The binary format string can be understood as:
     - "!" Use network byte order (big-endian).
+    - "I" store the entity ID as an unsigned int
     - "<encoding size>s" Store the n (e.g. 128) raw bytes of the bitarray
 
     https://docs.python.org/3/library/struct.html
     """
-    bit_packing_fmt = f"!{encoding_size}s"
+    bit_packing_fmt = f"!I{encoding_size}s"
```
hardbyte: We are just going to pretend the old format doesn't exist aren't we? Should we take the opportunity to add a version byte to the format?

backend/entityservice/serialization.py
```diff
@@ -54,31 +55,41 @@ def binary_pack_filters(filters, encoding_size):
...
 
 
-def binary_unpack_filters(streamable_data, max_bytes=None, encoding_size=None):
+def binary_unpack_filters(data_iterable, max_bytes=None, encoding_size=None):
+    """
+    Unpack filters that were packed with the 'binary_pack_filters' method.
+
+    :param data_iterable: an iterable of binary packed filters.
+    :param max_bytes: if present, only read up to 'max_bytes' bytes.
+    :param encoding_size: the encoding size of one filter, excluding the entity ID info
```
hardbyte: Since we bounce between bytes and bits I'd try be explicit about the size.

backend/entityservice/tasks/comparing.py
```diff
@@ -61,9 +61,8 @@ def create_comparison_jobs(project_id, run_id, parent_span=None):
         current_span.log_kv({"event": 'get-dataset-sizes'})
 
         filters_object_filenames = tuple(
-            get_filter_metadata(conn, dp_id) for dp_id in dp_ids)
+            get_filter_metadata(conn, dp_id)[0] for dp_id in dp_ids)
```
hardbyte: Might want to keep and use the  entity id indicies that `get_filter_metadata` is now returning?

backend/entityservice/tasks/comparing.py
```diff
@@ -125,10 +124,13 @@ def compute_filter_similarity(chunk_info, project_id, run_id, threshold, encodin
...
     log.debug("Fetching and deserializing chunk of filters for dataprovider 1")
     chunk_dp1, chunk_dp1_size = get_chunk_from_object_store(chunk_info_dp1, encoding_size)
-
+    #TODO: use the entity ids!
+    entity_ids_dp1, chunk_dp1 = zip(*chunk_dp1)
     t1 = time.time()
     log.debug("Fetching and deserializing chunk of filters for dataprovider 2")
     chunk_dp2, chunk_dp2_size = get_chunk_from_object_store(chunk_info_dp2, encoding_size)
+    # TODO: use the entity ids!
+    entity_ids_dp2, chunk_dp2 = zip(*chunk_dp2)
```
hardbyte: (what you said)

backend/entityservice/tasks/encoding_uploading.py
```diff
@@ -43,12 +43,12 @@ def handle_raw_upload(project_id, dp_id, receipt_token, parent_span=None):
     def filter_generator():
         log.debug("Deserializing json filters")
```
hardbyte: This log message looks  like a copy/paste error of mine no doubt. Could you update.

backend/entityservice/views/project.py
```diff
@@ -151,20 +151,26 @@ def project_binaryclks_post(project_id):
                 # connexion has already read the data before our handler is called!
                 # https://github.com/zalando/connexion/issues/592
                 # stream = get_stream()
-                stream = BytesIO(request.data)
-                expected_bytes = binary_format(size).size * count
+                stream = iterable_to_stream(BytesIO(request.data))
```
hardbyte: I'm confused. Isn't `BytesIO` already a stream? https://docs.python.org/3/library/io.html#binary-i-o

backend/entityservice/views/project.py
```diff
@@ -151,20 +151,26 @@ def project_binaryclks_post(project_id):
...
+                binary_formatter = binary_format(size)
+                def entity_id_injector(filter_stream):
+                    for entity_id in range(count):
+                        yield binary_formatter.pack(entity_id, filter_stream.read(size))
+                data_with_ids = b''.join(entity_id_injector(stream))
```
hardbyte: I want to check my understanding here. This takes the binary stream and converts it into a byte producing generator.
This generator uses the _user supplied_ `count` and reads `size` bytes (where `size` is also user supplied in the header) from the stream, packing the bytes using the binary format adding in an entity id as it goes.

- How do we handle the user supplying their own entity ids? I assume that is future work? Will they be optional/required?
- If this code path is here for the long haul, can we do this without making a full copy of the data in memory?

backend/entityservice/serialization.py
wilko77: our goal is 100 million. If there happens to be demand for more, we can fairly easily make adjustments.

backend/entityservice/serialization.py
wilko77: It's an internal format, so yes, I pretend that it never existed. I'm old, I forget quickly.
A version byte for each filter would mean a lot of version bytes in the object store. This should rather be solved via a header. But again, since this is only used internally in the service, we will never have to differentiate between the versions anyway and thus, the versioning is redundant.

backend/entityservice/serialization.py
wilko77: good point

backend/entityservice/tasks/comparing.py
wilko77: `get_filter_metadata` returns the filename and the encoding size of a dp's CLKs.

backend/entityservice/tasks/comparing.py
wilko77: yes, but this is another task. I want this PR to be single purpose.
Integrating the use of the entity IDs will most likely also require changes to the helper functions in the anonlink library. I didn't want to open that can of worm just yet.

backend/entityservice/serialization.py
hardbyte: yeah I agree

backend/entityservice/serialization.py
hardbyte: I was meaning a header for the file rather than for each encoding. But sure not worth it right now.

backend/entityservice/tasks/comparing.py
hardbyte: carry on then, nothing to see here

backend/entityservice/views/project.py
wilko77: yes it is. I got confused by the [read](https://docs.python.org/3.7/library/io.html?highlight=bufferedreader#io.BufferedIOBase.read) function, it says that it reads **up** to 'size' bytes. Thus I thought it might be saver to wrap it in a BufferedReader, but I guess that's unnecessary.

backend/entityservice/views/project.py
wilko77: I just wrote a quick fix to make this endpoint work with the new format.

Until connexion fixes the stream thing we cannot do much here.
There is also the problem, that the binary format as of now does not consider blocking information. 

I imagine, that in the far future, one day, someone will revisit the binary upload. (S)he will define a new binary format, implement that in the anonlink-client, and finally, will make this endpoint stunningly beautiful. 

For now, this is not a priority, though.

Issue #437, opened 2019-09-25
faster build of app container


hardbyte:
Our CI tests have a ~7 minute delay building the docker container `data61/anonlink-app` to test. A faster feedback cycle is desired.

Usually the dependencies don't change so this issue is to consider a base image that includes all our operating system dependencies and python requirements.

The basic idea is if a dependency changes we release a new version of `anonlink-base` and update our `anonlink-app` Dockerfiles to point to the new version.

hardbyte:
Some experiments in this branch - https://github.com/data61/anonlink-entity-service/compare/feature-separate-base-image

hardbyte:
One issue is CI testing changes to the base image - we'd like to regularly (automatically) update dependencies in the base image and have the CI system ensure they work. For the majority of builds where the base image doesn't change we don't want to wait the ~7 minutes for it.

One possible solution is to hash the Dockerfile + requirements file together (e.g. `sha256sum requirements.txt Dockerfile | sha256sum`) then use this hash as the Docker tag. The application docker image can then depend on the base using this hash, if it isn't present on Docker Hub the build must wait for the base build task, otherwise it should pull the correct version.

hardbyte:
Ugh this Azure pipeline gotcha got me - https://www.aaron-powell.com/posts/2019-05-24-azure-pipeline-templates-and-parameters/

hardbyte:
When it is working it is nice and fast:

![image](https://user-images.githubusercontent.com/855189/74687152-743ca880-5238-11ea-8c4a-ba634db0c458.png)

PR #506, opened 2020-02-18
Feature separate base image

hardbyte:
This PR addresses #437 by introducing a new layer of docker image for our "base" dependencies.

I'll clean up the commit history and remove the commented out azure debugging code but it would be good to get feedback.

After changing a dependency the building of the base image blocks the building off the app:

image

When the dependencies don't change the app builds nice and fast:

image


hardbyte:
(Rebased)

PR #511, opened 2020-02-20
Skip build in bash if azure isn't playing along

hardbyte:
The condition in azure didn't seem to take effect so we now check whether to skip the expensive image building in bash. I've tried various combinations of using strings and booleans in the main azure script and docker build template and cannot make Azure Pipelines do what I expect.

I updated the base digest too

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -88,7 +88,7 @@ stages:
       dependsOn: HashBaseDependencies
       imageName: data61/anonlink-app
       # this part is still todo automated
-      dockerBuildArgs: "--build-arg VERSION=02a973ce02c96d24376c5770c0426ccb0958a0d59136a4f2fc11c8666c43062a"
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -88,7 +88,7 @@ stages:
       dependsOn: HashBaseDependencies
       imageName: data61/anonlink-app
       # this part is still todo automated
-      dockerBuildArgs: "--build-arg VERSION=02a973ce02c96d24376c5770c0426ccb0958a0d59136a4f2fc11c8666c43062a"
+      dockerBuildArgs: "--build-arg VERSION=ed0be6ee130ec505b2a65af26ba570e1033ffb7e0cc6b27b6a8520df94f65d20"
```
wilko77: this is really annoying. When we change the base image requirements, although a new version gets build, it will not be tested. We already had a non-functioning base image. We really need to find a way to test the new base images straight away.

hardbyte:
Agree it could be nicer if Azure was a bit easier to use.

Have added the need to automate using and testing the latest base image to the backlog.

hardbyte:
Oh that is annoying, I set the merge target as `feature-blocking`. Will recreate...

PR #512, opened 2020-02-20
Feature clknblock some of the way

hardbyte:


backend/entityservice/tasks/encoding_uploading.py
```diff
@@ -30,69 +33,51 @@ def handle_raw_upload(project_id, dp_id, receipt_token, parent_span=None):
...
     # Output file is our custom binary packed file
     filename = Config.BIN_FILENAME_FMT.format(receipt_token)
-    bit_packed_element_size = binary_format(uploaded_encoding_size).size
-    num_bytes = expected_count * bit_packed_element_size
-
-    # If small enough preload the data into our redis cache
-    if expected_count < Config.ENTITY_CACHE_THRESHOLD:
-        log.info("Caching pickled clk data")
-        python_filters = list(python_filters)
-        encoding_cache.set_deserialized_filter(dp_id, python_filters)
```
hardbyte: We don't actually use this cache anywhere :-/

backend/entityservice/encoding_storage.py
```diff
@@ -0,0 +1,51 @@
...
+        Currently we do everything in memory.
+        Eventually we might want to instead stream through the input (perhaps twice).
+
+    :param f: File-like object containing `clksnblocks` json.
+    :return: a tuple comprising:
+        - dict mapping blocks to lists of bytes (each encoding in our internal encoding file format)
+        - the size of the first encoding in bytes (excluding entity ID info)
+    """
+    # Each block index contains a set of base64 encodings.
+    encodings_by_block = {}
```
wilko77: I don't think it is a good idea to store this in memory. If every encoding is in 10 different blocks, than  this structure is 10 times bigger than the already big json input `f`. I'd prefer if we would use a streaming solution that writes the encodings straight to file.

hardbyte:
Hey Wilko, I completely agree on both points.  My thought was get a minimum (barely) viable in-memory version that at least gets the `clknblock` format propagated deeper into the system to avoid blocking you for days while I'm not working. In any case I had started two branches ontop of this that start to do the 2 pass streaming approach that I wanted to do - I'll keep working on that this morning and then open a new (much larger) PR.

PR #513, opened 2020-02-24
Store encodings in postgres to support blocking

hardbyte:
This PR modifies the postgres database to store the raw encoding and blocking information.

I've modified the upload data task to insert the data into the database, but as this was already going to be a massive PR I have kept the existing object store file based approach in place. 

The comparison tasks have not been touched - they will have to be modified to pull data from the db instead of the object store.

The database now has these tables to store the encodings and blocks:
image


hardbyte:
Since this PR is based off commits that have been squashed and merged into develop I'll rebase.

PR #516, opened 2020-02-24
Store uploaded encodings in database

hardbyte:
This PR modifies the postgres database to store the raw encoding and blocking information.

I've modified the upload data task to insert the encoding data into the database, but have kept the existing object store file based approach in place. We now rewrite any user supplied JSON in the `clks` format into the `clksnblocks` format so our background task only has to deal with one format.

The comparison tasks have not been touched - they will have to be modified to pull data from the db instead of the object store.

The database now has these tables to store the encodings and blocks:

image


backend/entityservice/tasks/encoding_uploading.py
```diff
@@ -1,66 +1,76 @@
...
+
+    with opentracing.tracer.start_span('upload-encodings-to-db', child_of=parent_span):
+        # stream encodings with block ids from uploaded file
+        # convert each encoding to our internal binary format
+        # output into database for each block (temp or direct to minio?)
+        pipeline = convert_encodings_from_base64_to_binary(stream_json_clksnblocks(raw_data))
+        with DBConn() as db:
+            for entity_id, encoding_data, blocks in pipeline:
+                # write this encoding to files or database
+                insert_encoding_into_blocks(db, dp_id, blocks, entity_id, encoding_data)
```
hardbyte: It would be fair to say this should be improved. Doing insertions one at a time (in fact 2 SQL queries per encoding) is sub-optimal.

backend/entityservice/encoding_storage.py
```diff
@@ -0,0 +1,119 @@
...
+
+def store_encodings_in_db(conn, dp_id, encodings: Iterator[Tuple[str, bytes, List[str]]]):
+    """
+    Group encodings + blocks into database transactions and execute.
+
+    Assuming default encoding size of 128 B, n encodings each with their own
+    4 B encoding id, and assuming `k` multiple unique blocks of 64 B will be a transaction
+    of approximately k*64 + 132 * n. For k = 10 and n = 100_000 this gives a transaction
+    size under 100MiB.
+    """
```
wilko77: What happens if those assumptions don't hold? Let's say we run a comparison with wildly bigger encoding sizes. Will it all explode? Should we make `n` dependent on the encoding size?

backend/entityservice/init-db-schema.sql
```diff
@@ -128,19 +130,67 @@ CREATE TABLE bloomingdata (
...
+  -- Size in bytes of the uploaded encodings
   encoding_size  INT    NULL,
 
-  -- Number of uploaded entries
-  count  INT         NOT NULL
+  -- Number of uploaded encodings
+  count  INT         NOT NULL,
+
+  -- Number of blocks uploaded
+  block_count INT   NOT NULL DEFAULT 1
```
wilko77: both `count` and `block_count` are redundant info in the database. These values are also present int the `blocks` and `encodings` table.
We should be clearer on why we also need those values. What's their meaning?

backend/entityservice/database/selections.py
```diff
@@ -274,19 +275,42 @@ def get_bloomingdata_columns(db, dp_id, columns):
...
+
+    args = (dp_id, block_name) if block_name else (dp_id,)
+
+    cur.execute(sql_query, args)
+    while True:
+        rows = cur.fetchmany(10_000)
+        if not rows:
+            break
+        for row in rows:
+            yield row[0]
```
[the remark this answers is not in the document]
hardbyte: Why don't you think it would be good to keep the db connection while streaming through the blocks? Establishing a db connection is not free. The point of this sneaky Python cache is because we might not have the memory to store all (e.g. millions) of blocks if we used `fetchall()`, but the network overhead of `fetchone` if we fetched and yielded each row one at a time would be a killer.

backend/entityservice/init-db-schema.sql
hardbyte: They are a bit redundant - they can be computed by the database easily enough. 

I think the `count` of uploaded encodings is only used in scheduling/chunking the comparison tasks - the block_count won't be used anywhere yet (well it is notionally checked in the encoding_upload task).

PR #522, opened 2020-03-03
Use encodings from db instead of minio

hardbyte:
~Builds ontop of PR that fixes Jaeger: #523~

This PR completes the switch from storing encodings in minio to postgres. It makes several **short term assumptions** such as assuming there will be one *default block per data provider* and that encoding ids will be sequential non-repeating integers. These assumptions will be addressed over the next several sprints.

Chunking information previously included object store filenames, now each chunk comprises a  data provider ID and a range of *encoding ids*. E.g. `dp_id=45, range=[2000,5000]`, the comparison task now queries the database to fetch these encodings (currently ignoring blocks).

Tracing child spans was not working inside of celery tasks so I've fixed the tasks I touched.

### DIFF
--- a/backend/entityservice/integrationtests/dbtests/test_insertions.py
+++ b/backend/entityservice/integrationtests/dbtests/test_insertions.py
@@ -74,23 +74,14 @@ class TestInsertions:
         num_entities = 10_000
         blocks = [['1'] for _ in range(num_entities)]
         encodings = [data[i % 100] for i in range(num_entities)]
-        start_time = time.perf_counter()
         insert_encodings_into_blocks(conn, dp_id,
                                      block_ids=blocks,
                                      encoding_ids=list(range(num_entities)),
                                      encodings=encodings
                                      )
-        end_time = time.perf_counter()
-        elapsed_time = end_time - start_time
-        # This takes ~0.5s using docker compose on a ~5yo desktop.
-        # If the database is busy - e.g. if you're running integration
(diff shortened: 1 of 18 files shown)


backend/entityservice/integrationtests/dbtests/test_insertions.py
```diff
@@ -88,12 +88,14 @@ def test_insert_many_clks(self):
         assert elapsed_time < 2
 
         stored_encoding_ids = list(get_encodingblock_ids(conn, dp_id, '1'))
-        fetch_time = time.perf_counter() - end_time
+        id_fetch_time = time.perf_counter() - end_time
         # retrieval of encoding ids should be much faster than insertion
-        assert fetch_time < elapsed_time
-
+        assert id_fetch_time < elapsed_time
```
wilko77: these timing assertions might not always hold. (e.g. if there is a load spike in the db) So should we really fail if it goes wrong once?

backend/entityservice/views/project.py
```diff
@@ -184,25 +176,29 @@ def entity_id_injector(filter_stream):
...
-def project_clks_post(project_id):
-    """
-    Update a project to provide encoded PII data.
-    """
-    log = logger.bind(pid=project_id)
-    headers = request.headers
-
-    parent_span = g.flask_tracer.get_span()
-
+def precheck_encoding_upload(project_id, headers, parent_span):
```
wilko77: docstring would be nice

PR #527, opened 2020-03-09
Comparison tasks now based on blocks

hardbyte:
*Builds on from https://github.com/data61/anonlink-entity-service/pull/522 which moved encodings to the database*

This PR changes the comparison tasks to use blocking info. Where the individual blocks are too large we break them into "chunks".

This has a negative performance impact on small record linkage jobs. Many small blocks are compared slower than larger blocks. When the jobs get really massive though it is infeasible without blocking.


hardbyte:
Consider using `anonlink.candidate_generation._merge_similarities` instead for deduplication.

hardbyte:
During review this is deployed at https://blocked-anonlink.easd.data61.xyz/

backend/entityservice/tasks/comparing.py
```diff
@@ -1,81 +1,89 @@
...
     log = logger.bind(pid=project_id, run_id=run_id)
+    current_span = create_comparison_jobs.span
     with DBConn() as conn:
+        check_run_active(conn, project_id, run_id)
 
         dp_ids = get_dataprovider_ids(conn, project_id)
-        assert len(dp_ids) >= 2, "Expected at least 2 data providers"
-        log.info(f"Starting comparison of CLKs from data provider ids: "
+        number_of_datasets = len(dp_ids)
+        assert number_of_datasets >= 2, "Expected at least 2 data providers"
```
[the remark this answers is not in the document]
hardbyte: Oh good catch, yeah I wanted to reduce this duplication. Yup an assertion will lead to the celery task failing which will (now) mark the run as a failure due to the `link_error=run_failed_handler.s()` in run.py

hardbyte:
> 
> 
> you wrapped some logic around `anonlink.concurrency.split_to_chunks`. The alternative would be to modify/extend this functionality in anonlink itself. Did you consider that and what's your reasoning?

Most of the extra stuff relates to the database fields dataprovider id and block id which anonlink doesn't need to worry about. That said `_create_work_chunks` could, with some modification, find its way into the anonlink library at which point we would update the entity service to use the higher level functionality.

hardbyte:
I wasn't that happy with the time taken to retrieve the encodings from the database so I've changed to using a binary COPY command.

image

backend/entityservice/database/selections.py
```diff
@@ -275,51 +276,97 @@ def get_uploads_columns(db, dp_id, columns):
...
+    raw_data = stream.getvalue()
+
+    # Need to read/remove the Postgres Binary Header, Trailer, and the per tuple info
+    # https://www.postgresql.org/docs/current/sql-copy.html
+    ignored_header = raw_data[:15]
+    header_extension = raw_data[16:20]
+    assert header_extension == b'\x00\x00\x00\x00', "Need to implement skipping postgres binary header extension"
+    binary_trailer = raw_data[-2:]
+    assert binary_trailer == b'\xff\xff', "Corrupt COPY of binary data from postgres"
+    raw_data = raw_data[19:-2]
```
wilko77: This code to handle the Postgres binary format could be a helper function. We might need it again somewhere else. 
It would also clean up the code in here a bit, and make this function more single  purpose.

hardbyte: That's a good point, will do.

PR #531, opened 2020-03-24
Use local benchmark cache

hardbyte:
A small tweak to the benchmarking Docker image such that it supports mounting a volume to use as a download cache - this is to support developers using the NBN.


benchmarking/Dockerfile
```diff
@@ -15,10 +15,11 @@ RUN wget https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSI
 
 RUN python -c "import clkhash; print('clkhash version:', clkhash.__version__)"
 
+RUN mkdir /cache
 RUN adduser user && \
-    chown user:user /app /var/log
+    chown user:user /app /var/log /cache
+VOLUME /cache
```
hardbyte: This is required so that the `user` has permission to write to the mounted folder.

PR #533, opened 2020-03-24
Cleanup docker compose

hardbyte:


tools/docker-compose.yml
```diff
@@ -54,18 +58,6 @@ services:
     depends_on:
       - db
 
-  # The dm admin
-  db_admin:
```
wilko77: do we not need the db_admin any more?

base/requirements.txt
```diff
@@ -6,6 +6,7 @@ colorama==0.4.3 # required for structlog
 connexion==2.6.0
 Flask-Opentracing==1.1.0
 Flask==1.1.1
+flower==0.9.3
```
wilko77: what is that for?

hardbyte: they look pretty :rose:

wilko77: true

hardbyte: `flower` is a celery monitoring tool - we have an example use in the docker-compose file (commented out).

tools/docker-compose.yml
hardbyte: I suspect no one other than myself ever used it. It is useful to see the time queries take and see the locks on the database. Want me to put it in a `developer-tools.yml` file or something?

PR #536, opened 2020-04-04
Update tutorials to use anonlinkclient instead of clkhash

hardbyte:


backend/entityservice/settings.py
```diff
@@ -12,9 +12,6 @@ class Config(object):
     """
 
     # If adding or deleting any, please ensure that the changelog will mention them.
-
-    DEBUG = os.getenv("DEBUG", "false") == "true"
```
JoyceXinyueWang, line 16: Why is this not needed anymore?

hardbyte: We removed the debug flag a long time ago. Not sure why this hadn't been cleaned up

PR #537, opened 2020-04-06
Commit initial design of pull feature.

hardbyte:
Adds a design doc for a feature to "pull data" from an external object store.

If we are all happy with this design, I'll spend the rest of the week prototyping it and update the design doc accordingly. Unsure if we want to commit this kind of design document, or wait until the feature is implemented and just document the "actual" system.

## Background 

In early 2020 blocking has been added to increase the capacity of the Anonlink system, however the system cannot currently handle users uploading large encoded datasets in a reliable way. Investing in scaling Anonlink to a system that can deal with tens of millions of records vastly increases the chance of adoption and the general usefulness of the system.

When a client tries to upload tens of millions of records today they must have a fast and reliable internet connection and complete in one REST call the entire data upload operation. These uploads can easily be multiple GiB of data.

On the server side we have had to increase default timeouts to >60s to handle the upload of single digit millions of encoded records. The idea of the pull feature is to allow the client to point to a location where the data can be pulled by the server. Modifying the client and API to upload in chunks - making many smaller upload requests - was also considered. This alternative was rejected as we estimate it would take more effort to get to the same reliability. For example the chunking, compression, retrying, and data integrity would all have to be custom handled.

With the Anonlink system we aim to provide everything required to take users from clean records through to an accurate concordance table. We rejected the idea of requiring a data provider to setup their own object store or otherwise hosting their encoded data online; instead we will modify the client to be able to upload to an object store.


docs/image_source/data-upload.puml
```diff
@@ -0,0 +1,28 @@
...
+
+== Trigger data pull ==
+
+Alice -> Anonlink: Upload via ""/clks"" (optionally providing blocks)
+Anonlink --> Alice : Upload Receipt
+
+...some time later...
+
+Anonlink --> "Object Store" : Pull/copy encoding data
+
```
AJChapman, line 27: When are the temporary credentials revoked?

docs/designs/anonlink-upload-data-plan.md
```diff
@@ -0,0 +1,289 @@
...
+
+- A method for requesting temporary & restricted object store credentials shall be added to the Anonlink Entity Service 
+  API. The temporary credentials will authorize the holder to upload data, the credentials **must not** provide access 
+  to any other data providers encodings, or other working data stored in the object store.
+- A method for a data provider to provide an object store URI pointing to the encoding data instead of uploading it 
+  shall be added to the Anonlink Entity Service API. This method **may** support pulling directly via http as well.
+- A deployment option to expose the MinIO object store via an ingress will be added. As this increases the attack 
+  surface an internal security review of MinIO must be conducted. Deployment using default or publicly committed 
+  credentials must be mitigated.
+- The maximum size of a clients' encodings will be 1TiB.
```
AJChapman, line 52: s/clients'/client's (as we're talking about a singular client, adding the 's to indicate posession)

docs/designs/anonlink-upload-data-plan.md
```diff
@@ -0,0 +1,289 @@
...
+- High level system design
+- Specifications for implementation (including the REST API).
+- decisions and alternatives considered. What would be the "best"/"easiest" approach.
+
+
+## Requirements
+
+- A method for requesting temporary & restricted object store credentials shall be added to the Anonlink Entity Service 
+  API. The temporary credentials will authorize the holder to upload data, the credentials **must not** provide access 
+  to any other data providers encodings, or other working data stored in the object store.
```
AJChapman, line 46: s/providers/providers' (add an apostrophe at the end of this plural to indicate possession -- the data providers have encodings)

docs/designs/anonlink-upload-data-plan.md
```diff
@@ -0,0 +1,289 @@
...
+- Requirements
+- High level system design
+- Specifications for implementation (including the REST API).
+- decisions and alternatives considered. What would be the "best"/"easiest" approach.
+
+
+## Requirements
+
+- A method for requesting temporary & restricted object store credentials shall be added to the Anonlink Entity Service 
+  API. The temporary credentials will authorize the holder to upload data, the credentials **must not** provide access
```
AJChapman, line 45: s/authorize/authorise for Australian (NZ?) spelling, though this US spelling is probably ok in technical documentation

docs/image_source/data-upload.puml
```diff
@@ -0,0 +1,28 @@
...
+activate Anonlink
+
+Anonlink -> "Object Store": Create restricted credentials
+"Object Store" --> Anonlink : Credentials Response
+Anonlink --> Alice: Temporary object store credentials
+deactivate Anonlink
+
+== Upload Encodings ==
+
+Alice -> "Object Store": Upload Binary Encoding Data
```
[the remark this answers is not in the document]
hardbyte: Great point, I'll reference that in the **Abstract/Purpose** section of the design doc and add an **Alternatives** section.

The gist is we could modify the Anonlink api to deal with multipart uploads and add in the retrying mechanisms and integrity checks - and update our clients. There may be a lot of corner cases though and if we rely on an already solid object store client and server we should get to a reliable solution faster.

docs/designs/anonlink-upload-data-plan.md
```diff
@@ -0,0 +1,289 @@
...
+## High Level Design
+
+The Anonlink Entity Service already relies on MinIO (or AWS S3) as an ancillary services for storing encoding data
+as well as data for in progress computations. MinIO can easily and robustly handle the expected volume of data, and we 
+will leverage the object store to support this pull feature.
+
+At deployment time the user will have the option of exposing the MinIO object store used by the Anonlink Entity Service,
+or - as already supported - providing external S3/MinIO access credentials. The Anonlink Entity Service's API will be
+modified such that a data provider can request temporary object store credentials. The client uses these temporary
+credentials to upload their encodings to the object store. The client then informs the Anonlink server via the `clks` endpoint
```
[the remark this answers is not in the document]
hardbyte: That could work too. 

It could involve some trickery to create Minio users dynamically, the [docs](https://github.com/minio/minio/tree/master/docs/multi-user) only show creating users/groups via the `mc` command line tool and don't document the underlying API. I was thinking of creating a default upload user using `mc` at service startup, and then using the assume role api to create credentials for each data uploader who requests them.

The [AssumeRole](https://github.com/minio/minio/blob/master/docs/sts/assume-role.md) API only creates temporary credentials (valid between 15min to 12 hours).

From a user point of view I don't see it making a big difference, the majority of users will be using our client tool which can happily make the three calls in sequence. Also this is only for the case where we are exposing our object store, if the user already has object store credentials e.g. for S3 they can upload at their leisure without having to think about temporary credentials. It would be nice for our client tool to deal with this nicely/transparently.

The default case using temporary credentials for our object store:
``` 
$ anonlink upload mydata.csv <AUTH params etc>
```

Where the user wants our client to upload to their own bucket, optionally providing the AWS credential profile to use:
```
$ anonlink upload mydata.csv [--profile easd] --upload-to=s3://my-own-bucket/mydata.csv <AUTH params etc>
```

If the user wants to use already uploaded data, they will have to provide the anonlink entity service with appropriate credentials. I definitely don't want to assume we can share a users AWS credentials with the service, and I'm not sure if we should be asking the users' AWS account for (temporary) credentials, or just require it to be all explicitly provided.

This means something like:
```
$ anonlink upload s3://my-own-bucket/mydata.csv
```
is probably not going to work unless the user has provided us explicit credentials to share with the service, or explicitly told us that our client should create temporary and restricted credentials to share with the service. Telling us could be additional command line arguments, an `~/.anonlink` config file or environment variables. 

cc @wilko77 do you have opinions here?

docs/image_source/data-upload.puml
hardbyte: Probably you got this from my earlier answer but they time out.

docs/designs/anonlink-upload-data-plan.md
```diff
@@ -0,0 +1,424 @@
...
+
+```sh
+$ anonlink upload s3://my-own-bucket/mydata.csv [--profile easd] --request-read-only-credentials \
+           <AUTH params etc>
+```
+
+or 
+```sh
+$ anonlink upload s3://my-own-bucket/mydata.csv [--profile easd] --share-aws-credentials-with-server \
+           <AUTH params etc>
```
wilko77, line 333: This are a lot of different ways to upload data, all under the same command. That might lead to a long list of conditionals in the code, also the docs for the command will be rather long to cover all those cases. 
How about we split the cases where we link to already uploaded data and put them into a different command?

hardbyte: That's a good idea, I've added `anonlink serverpull` for that second use case. I'm thinking it might be a stretch to implement that in the first revision though.

wilko77: that's fine. I think we should aim to implement the most common use case and it's probably fine to leave the others in the planning stage.

PR #543, opened 2020-04-17
Fix calculation of total number of comparisons

### LINKED ISSUE #542: Progress metrics not taking blocks into account
The run progress calculation isn't taking into account the size of blocks.

hardbyte:
This calculation now takes into account the use of blocking. Also because it seemed somewhat non-trivial to compute I added caching in redis.

Closes #542

### DIFF
--- a/backend/entityservice/tasks/run.py
+++ b/backend/entityservice/tasks/run.py
@@ -45,7 +45,6 @@ def prerun_check(project_id, run_id, parent_span=None):
 
         log.debug("Updating redis cache for run")
         set_run_state_active(run_id)
-        progress_cache.save_current_progress(comparisons=0, run_id=run_id)
 
     create_comparison_jobs.apply_async(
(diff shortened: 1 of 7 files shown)


backend/entityservice/tasks/run.py
```diff
@@ -45,7 +45,6 @@ def prerun_check(project_id, run_id, parent_span=None):
 
         log.debug("Updating redis cache for run")
         set_run_state_active(run_id)
-        progress_cache.save_current_progress(comparisons=0, run_id=run_id)
```
hardbyte, line 48: This was why I'd seen a log messages saving "0 comparisons".

hardbyte:
Yeah it was just a dumb 2 party assumption on my side, I was first calculating the total number of comparisons as:

> For each data provider multiple the size of each block together, then sum resulting number
    of comparisons in each block all together.

When the actual calculation needed to be:

```
for each block:
        for each **pairwise combination** of data providers in each block:
                multiply the block sizes together
        then sum number of comparisons for each block
Sum the number of comparisons in all blocks together
```

PR #544, opened 2020-04-20
Temporary object store credentials endpoint

hardbyte:
This PR adds an endpoint to retrieve temporary object store credentials.

The design is documented [here](https://github.com/data61/anonlink-entity-service/blob/develop/docs/designs/anonlink-upload-data-plan.md#addition-of-an-endpoint-to-retrieve-temporary-object-store-credentials). The main thrust of this is complete, but I still haven't finished the k8s deployment side of things. In particular I'd like to get the k8s initialization working and remove any object store credentials before adding support to open the minio object store to the internet.


backend/entityservice/views/objectstore.py
```diff
@@ -0,0 +1,79 @@
...
+          "Sid": "Upload-access-to-specific-bucket-only"
+        }
+      ]
+    }
+
+    return json.dumps(restricted_upload_policy)
+
+
+def authorize_external_upload(project_id):
+    assert config.UPLOAD_OBJECT_STORE_ENABLED, "Feature disabled"
```
wilko77, line 37: will that create an appropriate response for the request?

tools/ci.yml
```diff
@@ -16,14 +15,19 @@ services:
     image: data61/anonlink-app:${TAG:-latest}
     environment:
       - SERVER=http://nginx:8851
-      - INITIAL_DELAY=20
```
wilko77, line 19: you removed the env var `INITIAL_DELAY` in a couple of places. Who uses this? What is it defaulting to now?

backend/entityservice/api_def/openapi.yaml
```diff
@@ -1076,3 +1104,32 @@ components:
...
+          description: Object Store credentials (compatible with both AWS & MinIO)
+          type: object
+          properties:
+            AccessKeyId:
+              type: string
+            SecretAccessKey:
+              type: string
+            Expiration:
+              type: string
+            SessionToken:
```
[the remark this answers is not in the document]
hardbyte: Exactly the same - thankfully

backend/entityservice/views/objectstore.py
hardbyte: I checked that the endpoint returns a `500` and the schema is correctly a `Problem`. However it doesn't actually send the "feature disabled" message to the user. I'll look into it.

tools/ci.yml
hardbyte: It is an optional environment variable used by both the integration tests and end to end tests - I probably shouldn't have deleted them as they serve as documentation.

Note I also added a small sleep in the docker `command` for the integrationtests container

backend/entityservice/views/objectstore.py
hardbyte: I forgot we had a `send_fail_request` for this exact purpose - fixed now.

PR #547, opened 2020-04-21
Use docker-compose without having to set the --project-directory

hardbyte:
Just a small tidy up, there was some inconsistency between how we used docker-compose and how the average user might expect to.


.azurePipeline/runDockerComposeTests.sh
```diff
@@ -56,11 +56,14 @@ process_args "$@"
...
 export TAG=$TAG
 
-commandPrefix="docker-compose -f tools/docker-compose.yml -f tools/ci.yml --project-directory . "
+commandPrefix="docker-compose -f docker-compose.yml -f ci.yml "
 if [[ "$NO_ANSI" == "TRUE" ]]; then
   commandPrefix="$commandPrefix --no-ansi "
 fi
 
+# change current working directory to the tools directory
+cd "${0%/*}/../tools"
```
wilko77, line 65: yeah right. I'm sure there is a universe where that command makes intuitive sense...

PR #548, opened 2020-04-22
Expose minio

hardbyte:
This PR is mostly around the kubernetes deployment of minio it builds on #547 .

- removes default minio credentials 
- provides info through the temp auth api about whether minio requires a `secure` connection
- tweaks the temp auth api to return a 201 rather than 200 status code.
- uses the external address of minio, if the user has configured an ingress
- I updated the notes that are printed on deployment too.

Other than the e2e tests I've also [drafted a branch of anonlink-client](https://github.com/data61/anonlink-client/pull/46) which uses a [deployment of this expose-minio branch](https://blocked.anonlink.easd.data61.xyz/) to fetch credentials and upload data directly to the object store.


backend/entityservice/object_store.py
```diff
@@ -30,7 +30,7 @@ def connect_to_upload_object_store():
         config.UPLOAD_OBJECT_STORE_ACCESS_KEY,
         config.UPLOAD_OBJECT_STORE_SECRET_KEY,
         region="us-east-1",
```
wilko77, line 32: why is region hard-coded when everything else is configurable?

deployment/entity-service/values.yaml
```diff
@@ -9,14 +9,24 @@ anonlink:
   config: {}
 
   objectstore:
-    uploadEnabled: "true"
-    # TODO can't leave defaults once minio exposed
+    ## Disable the feature providing client's with restricted upload access to the object store.
+    ## By default we don't expose the Minio object store, however it needs to be exposed when
+    ## enabling this feature. See section `minio.ingress`.
```
wilko77, line 14: I don't understand this comment. So what do I have to set this to to enable client upload to our Minio object store?

backend/entityservice/object_store.py
hardbyte: It could be configurable but we'd probably have to introduce another minio/boto client just to connect to `us-east-1` for the granting the temporary credentials.

deployment/entity-service/values.yaml
hardbyte: I'll clarify this

PR #551, opened 2020-04-27
Feature external blocks

hardbyte:
Adds backend implementation of pulling blocking information from external object store.

As we haven't defined a binary format for blocking information I made the decision to stick with the JSON format, but simply process it from an external file.


hardbyte:
This is a draft that is close to working but still needs to be rebased

PR #563, opened 2020-05-26
Mini blocks speed

wilko77:
this PR is mainly about handling lots of very small blocks.
This is important, as the p-sig blocking aims to create lots of very small blocks.
Previously, we created a task for each block. This is not very efficient, as the block usually contains some single digit numbers of encodings. Now, we create work packages for the tasks, which can contain many blocks, but are also backwards compatible, and thus, can still chunk too large blocks into several packages.

In term of db, we had some weird performance problems. As far as I could figure out, it had to do with indices. Whenever the database created those indices, the performance was very slow. I re-shuffled the db schema, such that we now have block_ids, which are integers, and block_names, which are user-provided block identifier. And encoding_id, which are unique integers in the db to identify a encoding, and entity_ids, which are user provided.


backend/entityservice/database/insertions.py
```diff
@@ -89,23 +90,33 @@ def insert_encodings_into_blocks(db, dp_id: int, block_ids: List[List[str]], enc
...
+    ## first we have to look up the block numbers corresponding to the given block_names
+    block_info_iter = get_block_metadata(db, dp_id)
+    block_lookup = {bl_name: bl_id for bl_name, bl_id, _ in block_info_iter}
+
+    encodings_insertion_query = "INSERT INTO encodings (encoding_id, encoding) VALUES %s"
+    blocks_insertion_query = "INSERT INTO encodingblocks (dp, entity_id, encoding_id, block_id) VALUES %s"
+    # we differentiate between entity_id and encoding_id.
+    # The entity_id is the id that the dataprovider assigns to an entity. Usually the row number of that entity in the
+    # dataset.
+    # The encoding_id is used internally to address encodings uniquely.
```
hardbyte, line 102: Is the intention that there will be multiple `encoding-ids` per `entity-id`? Or is the `encoding_id` just a primary key for the encodings table?

I suggest renaming one or both to make it clearer.  Perhaps `encoding_pk` instead of `encoding_id`.

backend/entityservice/database/util.py
```diff
@@ -145,3 +147,41 @@ def execute_returning_id(cur, query, args):
...
+            yield row_values
+
+
+def compute_encoding_ids(entity_ids: Iterable[int], dp_id: int):
+    """ compute unique encoding ids for given entity ids
+    The user provides entity ids. Although unique for the user, different user can have the same entity ids.
+    However, we want to use the id as a unique identifier for the encoding in the database. Thus, we create a
+    unique number here. """
+    dp_offset = dp_id << 32
+    return [dp_offset + entity_id for entity_id in entity_ids]
```
hardbyte, line 187: Does the `entity_id` need to be encoded into the `encoding_id`? If not perhaps it would be better to simply insert the encodings and have the database assign unique row keys?

backend/entityservice/tasks/comparing.py
```diff
@@ -126,34 +122,60 @@ def _create_work_chunks(blocks, dp_block_sizes, dp_ids, log, chunk_size_aim=Conf
...
+            num_comparisons = size1 * size2
+            return block_id, dp1, dp2, size1, size2, num_comparisons
+        except StopIteration as e:
+            raise e
+
+    block_name, dp1, dp2, size1, size2, comparisons = next_block(blocks)
+    cur_package = []
+    cur_comparisons = 0
+    try:
+        while True:
```
hardbyte, line 142: Why not iterate directly over `next_block` (after converting it to a generator)? Then you could avoid using exceptions for control flow

backend/entityservice/tasks/comparing.py
```diff
@@ -241,91 +292,132 @@ def new_child_span(name, parent_scope=None):
...
+    #chunk_info_dp1, chunk_info_dp2 = chunk_info
+    def reindex_using_encoding_ids(recordarray, encoding_id_list):
+        # Map results from "index in chunk" to encoding id.
+        return array.array('I', [encoding_id_list[i] for i in recordarray])
+
+    num_results = 0
+    num_comparisons = 0
+    sim_results = []
+
+    if len(package) > 1:  # multiple full blocks in one package
```
hardbyte, line 308: I'm surprised to see this isn't the only path - why not just iterate over the packages knowing that sometimes there is only one package (when the whole chunk is part of the same block)?

backend/entityservice/database/util.py
wilko77: If you let the db assign unique ids, we run into a problem: First some task inserts the encodings. This is fine. At some later stage, another task wants to grab some of the encodings. This tasks does not know the unique ids, thus has to query them with 'encoding id' and 'dp id'. To speed this up, we would have to create an index for (encoding_id, dp_id) on the encodingblocks table in the database. In my tests, this index creation was the main culprit for the mediocre performance. Once created, everything's nice and fast, but during creation everything grinds to a halt.  
I couldn't figure out how to speed up the index creation. That's why I introduced this unique ids to get around the index.

hardbyte: > This tasks does not know the unique ids, thus has to query them with 'encoding id' and 'dp id'. To speed this up, we would have to create an index for (encoding_id, dp_id) on the encodingblocks table in the database.

Ahh right I'm with you, I'd forgotten we now might be selecting encodings by multiple blocks at once or as a chunk within one block.

> In my tests, this index creation was the main culprit for the mediocre performance. Once created, everything's nice and fast, but during creation everything grinds to a halt.
I couldn't figure out how to speed up the index creation.

That's a shame, I'd have thought it could be done but might not be worth the added complexity. E.g. inserting into a temporary table without an index, adding the index and then copying into the real table. 

I guess the main alternative is a table mapping encoding_ids to a work package?

Anyhow the unique ids make more sense now

PR #564, opened 2020-06-01
Fix upload to external store

JoyceXinyueWang:
This pull request is to fix some bugs arise when testing encoding and blocks uploads to external object store.

1. The credentials used to create bucket should be environment credentials with MINIO_ACCESS_KEY and MINIO_SECRET_KEY rather than credentials from clients since it causes access denied error.

2. The uploaded encodings are not actual bytes but encoded strings. We need to deserialise the encodings properly before appending entity_id which assumes encodings are bytes stream

3. The uploaded blocks has two elements - blocks and metadata. Pull blocks only when writing to database.

Please review this after merging #563


backend/entityservice/tasks/encoding_uploading.py
```diff
@@ -56,12 +60,21 @@ def pull_external_data(project_id, dp_id,
...
 
+    # load clks properly
+    encodings = json.loads(encodings_stream.data.decode())['clks']
+
+    stream = io.BytesIO()
+    for clk in encodings:
+        stream.write(deserialize_bytes(clk))
+    stream.seek(0)
+    encodings_stream = stream
+
```
wilko77, line 77: it seems a bit harsh to remove the support for binary uploading completely. Better: support both.

backend/entityservice/tasks/encoding_uploading.py
```diff
@@ -56,12 +60,21 @@ def pull_external_data(project_id, dp_id,
...
 
-    stat, encodings_stream = stat_and_stream_object(bucket_name, object_name, parse_minio_credentials(encoding_credentials))
+    stat, encodings_stream = stat_and_stream_object(bucket_name, object_name, env_credentials)
     count = int(stat.metadata['X-Amz-Meta-Hash-Count'])
     size = int(stat.metadata['X-Amz-Meta-Hash-Size'])
     log.debug(f"Processing {count} encodings of size {size}")
     assert count == len(encoding_to_block_map), f"Expected {count} encodings in blocks got {len(encoding_to_block_map)}"
 
+    # load clks properly
+    encodings = json.loads(encodings_stream.data.decode())['clks']
```
wilko77, line 70: consider using `ijson`. It allows you to iteratively read json and thus avoids having to load the whole file into memory.

backend/entityservice/tasks/encoding_uploading.py
wilko77, line 77: Previously, we grabbed one encoding at a time from the encodings_stream and wrote it straight to the database. Enriching it in between with a generator. This is very memory efficient, as it never has to hold the whole file in memory. 
With your approach, you first create a full json representation in memory, and then a second one as BytesIO. That is quite wasteful. Consider uploads of 100s of millions of entities. That's quite a few gigabytes of data!

PR #582, opened 2020-08-11
update python3 version

wilko77:
turns out the base image was broken.
This PR fixes this.


wilko77:
Now the problem is the anonlink-client. Version 0.1.3 has a broken setup.py file. We really need to release a fixed version.

PR #587, opened 2020-12-01
Update ingress

hardbyte:
- Allows user to specify a path e.g. a prefix.
- remove default ingress config from values file (as users have to override them all)
- Adds a note to the documentation around ingress configuration with example for nginx.

~The pipeline appears to be failing (here and on [develop](https://dev.azure.com/data61/Anonlink/_build/results?buildId=2948&view=results)) due to this recently fixed issue:
https://github.com/microsoft/azure-pipelines-tasks/issues/13911~

### DIFF
--- a/deployment/entity-service/templates/api-ingress.yaml
+++ b/deployment/entity-service/templates/api-ingress.yaml
@@ -17,7 +17,7 @@ spec:
   - host: {{ . }}
     http:
       paths:
-      - path: /
+      - path: {{ .Values.api.ingress.path }}
         backend:
           serviceName: {{ $serviceName }}
           servicePort: {{ $servicePort }}
--- a/deployment/entity-service/values.yaml
+++ b/deployment/entity-service/values.yaml
@@ -122,36 +122,38 @@ api:
     enabled: false
 
     ## Ingress annotations
-    annotations: # {}
-     kubernetes.io/ingress.class: nginx
+    annotations: {}
+      ## Suggested annotations
+      ## To handle large uploads we increase the proxy buffer size
+      #ingress.kubernetes.io/proxy-body-size: 4096m
 
-     ## To handle large uploads we increase the proxy buffer size
-     ingress.kubernetes.io/proxy-body-size: 4096m
-     nginx.ingress.kubernetes.io/proxy-body-size: 4096m
+      ## Redirect to ssl
+      #ingress.kubernetes.io/force-ssl-redirect: "true"
 
-     nginx.ingress.kubernetes.io/proxy-connect-timeout: "60"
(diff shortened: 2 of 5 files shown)


deployment/entity-service/templates/api-ingress.yaml
```diff
@@ -17,10 +21,19 @@ spec:
   - host: {{ . }}
     http:
       paths:
-      - path: /
+      - path: {{ .Values.api.ingress.path }}
+        pathType: {{ .Values.api.ingress.pathType }}
```
hardbyte, line 25: Oh this isn't quite correct. This was added to k8s at some point.

In [version 1.16 of the k8s reference docs](https://v1-16.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/#httpingresspath-v1beta1-extensions) `pathType` isn't part of `apiVersion: extensions/v1beta1`. It is part of the [1.19 reference docs](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#httpingresspath-v1beta1-extensions) for ingress `apiVersion: extensions/v1beta1`.

I don't think this project offers any guidance as to the supported version of kubernetes but I think this might fail on the CI cluster.

deployment/entity-service/values.yaml
```diff
@@ -118,36 +118,38 @@ api:
...
+      ## https://kubernetes.io/docs/concepts/services-networking/ingress/#deprecated-annotation
+      # kubernetes.io/ingress.class: ""
+
+    path: /
+    pathType: Prefix
 
     ## Entity Service API Ingress hostnames
     ## Must be provided if Ingress is enabled
-    hosts:
-     - beta.anonlink.data61.xyz
```
hardbyte, line 139: @wilko77 I suggest these removed/commented out values might go into a private data61 deployment repo

wilko77: Wasn't there a problem with uploads and timeouts? Are those nginx timeouts essential for the ES to work? If so, we should ensure that they are either set to sensible defaults, or there are big hints in the doc that these values need to be set.

hardbyte: Ok I'll document the timeout issues we ran into and the workaround we used for nginx ingress controllers.

The issue is that nginx is just one of many options for a Kubernetes cluster's "ingress controller".  These settings were correct for deployments to the Data61 development/demo cluster but the chart's values.yaml are defaults for everyone.

hardbyte:
I'll rebase this after #596 is merged in (just got to fix the docs bulider)

wilko77 (approved):
lgtm

PR #590, opened 2020-12-04
Fix case sensitivity in minio metadata

hardbyte:
Field names in HTTP headers are case-insensitive, some networks decide that means that can normalize the headers however they like, presumably to keep things interesting.

On one such network the backend errors with the following:
```
Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/celery/app/trace.py", line 385, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/var/www/entityservice/tasks/base_task.py", line 87, in __call__
    return super(TracedTask, self).__call__(*args, **kwargs)
  File "/usr/lib/python3.8/site-packages/celery/app/trace.py", line 650, in __protected_call__
    return self.run(*args, **kwargs)
  File "/var/www/entityservice/tasks/encoding_uploading.py", line 72, in pull_external_data
    count = int(stat.metadata['X-Amz-Meta-Hash-Count'])
KeyError: 'X-Amz-Meta-Hash-Count'
```

Minio's `stat.metadata` is a dict of custom HTTP headers used to transfer metadata from the anonlink-client to the server via the object store. This small change ensures that the server will get the correct values regardless of the case of the header.

### DIFF
--- a/backend/entityservice/tasks/encoding_uploading.py
+++ b/backend/entityservice/tasks/encoding_uploading.py
@@ -1,5 +1,6 @@
 import json
 import ijson
+from requests.structures import CaseInsensitiveDict
 
 from entityservice.database import *
 from entityservice.encoding_storage import stream_json_clksnblocks, convert_encodings_from_base64_to_binary, \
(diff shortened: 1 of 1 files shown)


joyceyuu (approved):
Look good to me!

PR #594, opened 2020-12-18
Feature to expose object store similarities

hardbyte:
Building on #591 what do you think of exposing the filename of the similarity scores? One of the weakest links is the transferring of large amounts of data through the Python REST apis in single requests, the similarity scores are one of the biggest files the service creates.

We already have the data in an object store, so it is trivial to reply with the filename instead of the data. This PR simply checks for a header, but of course could be a new endpoint instead. Thoughts?


wilko77:
In general, I think this is a good idea. The only concern I have is with authorization. 
We must ensure that only people who have the result token can download the similarity scores.
Obviously, if you have the token, you will get the correct download URL. 
However, what if you don't have the token. Can you maybe guess the URL? Did you consider that in your approach?
Here's an idea: Can we use the pre-signed url functionality of minio? So you hit the endpoint with the token, then the endpoint generates the pre-signed URL for you. This way, we can rely on pre-sign mechanism to ensure that no un-authorized person can access the file.

hardbyte:
> In general, I think this is a good idea. The only concern I have is with authorization.
> We must ensure that only people who have the result token can download the similarity scores.
> Obviously, if you have the token, you will get the correct download URL.
> However, what if you don't have the token. Can you maybe guess the URL? Did you consider that in your approach?
> Here's an idea: Can we use the pre-signed url functionality of minio? So you hit the endpoint with the token, then the endpoint generates the pre-signed URL for you. This way, we can rely on pre-sign mechanism to ensure that no un-authorized person can access the file.

There is a bit of security in depth here. For this to work the caller must be able to connect to the object store (it isn't exposed in the default deployment). Also the caller having at least read access to the similarity scores bucket in the object store is a requirement for this feature to be helpful (as currently proposed). 

I agree only a caller with the result token should get to learn the resulting file name. I understand your concern that the caller might then be able to access other results by guessing the object store paths, however I'm not sure there is any vulnerability. The S3 api won't reveal if a file is actually present except to authenticated users, and authenticated users can just list the files.

FYI: The intermediate files during comparisons are at a well known address (e.g. `s3://entityservice/similarity-scores/RANDOM-FILENAME`) with a 12 byte (96bit) random file name. Knowing another results file name doesn't help an attacker at all as they won't have access to the object.

The caller doesn't gain access by learning the object store path, it just makes it easier for them to pull the data **if** they have access to the object store. Creating temporary access credentials valid to pull the file for a limited would be a good approach to provide the caller with restricted access (noting they may already have access) - I'll add this to the PR.

In the default deployment using MinIO the system uses root credentials, with the exception of generating temporary upload credentials - longer term this could be improved.

Another idea is to have the system copy the data to a different object store using caller provided credentials. That would put the responsibility back on the caller for providing the system with secure access to an object store where they can then pull the data. This is an extra copy of a large amount of data though, and would probably need a back end task to carry out the copy, monitoring/notification to know when it is complete etc.

hardbyte:
Oh I forgot about this, I changed it to use the pre-signed url functionality of minio. If you're happy I'll add a line or two of documentation

backend/entityservice/views/run/results.py
```diff
@@ -46,9 +48,15 @@ def get_result(dbinstance, project_id, run_id, token):
         return {"groups": result}
 
     elif result_type == 'similarity_scores':
-        logger.info("Similarity result being returned")
-        return get_similarity_score_result(dbinstance, run_id)
-
+        if 'INTERNAL_OBJECT_STORE_ADDRESS' in request.headers:
```
wilko77, line 51: I wonder if we can find a more descriptive header. Maybe `RETURN_OBJECT_STORE_ADDRESS`?

wilko77 (commented):
looks good to me. Needs testing, though. And some docu.

backend/entityservice/views/run/results.py
hardbyte: Sure that would be clearer, or would you think a a separate rest endpoint would be better?

hardbyte:
After https://github.com/data61/anonlink-entity-service/pull/601 I'll rebase and keep working on this.

tools/ci.yml
```diff
@@ -32,6 +26,21 @@ services:
       - nginx
       - objectstore_init
 
+  currenttest:
```
wilko77, line 29: what are current tests?

backend/entityservice/api_def/openapi.yaml
```diff
@@ -672,6 +677,38 @@ paths:
...
+        ```
+
+        The file itself is serialized using `anonlink.serialization`, you can convert the stream into Python types
+        with:
+
+        ```python
+        mc = Minio(file_info['endpoint'], ...)
+        candidate_pair_stream = mc.get_object(file_info['bucket'], file_info['path'])
+        sims, (dset_is0, dset_is1), (rec_is0, rec_is1) = anonlink.serialization.load_candidate_pairs(candidate_pair_stream)
+        ```
```
wilko77, line 710: nice

tools/ci.yml
```diff
@@ -14,14 +13,9 @@ services:
 
   integrationtests:
     image: data61/anonlink-app:${TAG:-latest}
+    env_file:
+      - .env
```
wilko77, line 17: sweet

wilko77 (approved):
looks good.

tools/ci.yml
hardbyte: Oh, pushing that was my mistake. Just a container that only runs the test I was interested in

hardbyte:
I'm still on the fence regarding a dedicated endpoint versus further overloading the `/results` endpoint.
Do you think there should be a feature flag to globally disable this?

I've also assumed that the data won't need copying - which of course won't be true if the user has configured separate download and internal object stores.

wilko77:
Instead of creating yet another way to interact, I would remove the old way of returning similarity scores. It never really worked for anything else but toy examples.
This is a potential source of bad user experiences, as they don't necessarily know what the internal limits of the server are. It also seems to be more straining for the server, and thus won't scale as neatly as your new way of doing things.

PR #596, opened 2020-12-20
Migrate off deprecated K8s dependencies

### LINKED ISSUE #588: Migrate off deprecated K8s dependencies
See https://github.com/helm/charts/tree/master/stable/redis-ha#%EF%B8%8F-repo-archive-notice

Both Redis and Minio are affected
- redis-ha has moved to https://github.com/DandyDeveloper/charts/tree/master/charts/redis-ha
- minio have taken control of their own helm chart at https://helm.min.io/

hardbyte:
This PR updates Minio and Redis helm dependencies. Note both charts have moved from the `stable` repository.

- redis-ha has moved to https://github.com/DandyDeveloper/charts/tree/master/charts/redis-ha
- minio have taken control of their own helm chart at https://helm.min.io/

This also updates the version of alpine used in the base image, and gets azure comfortable using helm v3.

The biggest change is probably the behaviour of the "external minio server" configured by `UPLOAD_OBJECT_STORE_SERVER`. I had made an assumption that the provisioned Minio would either be exposed via an ingress, or an external object store would be used. There are many cases where the system might be exposed differently so that is now a bit more configurable.

Replaces #591

Closes #588

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -270,15 +270,7 @@ stages:
       # Install Helm on an agent machine
       - task: HelmInstaller@1
         inputs:
-          helmVersionToInstall: '2.16.3'
-
-      - task: HelmDeploy@0
-        displayName: Helm init
-        inputs:
-          connectionType: 'Kubernetes Service Connection'
-          kubernetesServiceConnection: $(CLUSTER)
-          command: init
-          arguments: --client-only
+          helmVersionToInstall: 'latest'
 
       - task: HelmDeploy@0
         displayName: Helm package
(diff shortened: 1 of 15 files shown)


hardbyte:
From the other PR:

I pinned `bitarray-hardbyte==1.2.2` because of a test failure with the newer 1.6 release.
I also tried to update the `minio-py` client to `6.0.2` to be compatible with `anonlink-client`... but went down the rabbit hole and found you'd already been there.

wilko77 (approved):
It would be extremely nice if you could describe the object store options in the docs as well.
It seems that there are now quite a few, but little guidance for a newbie.

Otherwise, great.

PR #600, opened 2021-02-09
Update base

hardbyte:
Takes over from dependabot on #599 by updating all the alpine things.

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.12.1
+FROM alpine:3.13.1
 
 ENV DOCKERIZE_VERSION v0.6.1
 RUN wget https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSION/dockerize-alpine-linux-amd64-$DOCKERIZE_VERSION.tar.gz \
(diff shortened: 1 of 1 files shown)


hardbyte:
Gah, I forgot this uses a master and develop branch. Let me rebase and force push...

hardbyte:
Well [it looks like](https://dev.azure.com/data61/Anonlink/_build/results?buildId=3337&view=logs&jobId=9587a1dd-b2a7-5ab3-8262-37bc4f6ffd74&j=9587a1dd-b2a7-5ab3-8262-37bc4f6ffd74&t=686d1ff3-68fb-5ae5-5ef3-19f5f0b0833e) the current data61 libraries are incompatible with each other:

```
Collecting blocklib>=0.1.4
  Downloading blocklib-0.1.7-py3-none-any.whl (22 kB)
INFO: pip is looking at multiple versions of anonlink to determine which version is compatible with other requirements. This could take a while.
ERROR: Cannot install -r requirements.txt (line 2) and minio==5.0.10 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested minio==5.0.10
    anonlink-client 0.1.4 depends on minio>=6.0.0

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/user_guide/#fixing-conflicting-dependencies
```

wilko77:
why does the base image contain anonlink-client?

hardbyte:
It is imported by the benchmark script, but of course the benchmark image could add it directly.

Issue #133, opened 2018-04-18
Replace low level database operations with an ORM


hardbyte:
I quite like the look of [peewee](http://peewee.readthedocs.io/en/latest/index.html) although SqlAlchemy is the standard.

For example we would define `models` for our tables:

```
class Project(BaseModel):
    access_token = pw.TextField()
    chunk_size = pw.BigIntegerField(constraints=[pw.SQL("DEFAULT '-1'::integer")])
    notes = pw.TextField(null=True)
    parties = pw.IntegerField(constraints=[pw.SQL("DEFAULT 2")], null=True)
    ready = pw.BooleanField(constraints=[pw.SQL("DEFAULT false")])
    resource = pw.CharField(column_name='resource_id', unique=True)
    result_type = UnknownField()  # Or use USER-DEFINED
    schema = BinaryJSONField()
    threshold = pw.FloatField()
    time_added = pw.DateTimeField(constraints=[pw.SQL("DEFAULT CURRENT_TIMESTAMP")], null=True)
    time_completed = pw.DateTimeField(null=True)
    time_started = pw.DateTimeField(null=True)

    class Meta:
        table_name = 'projects'
```

We can then create tables:

```
def create_tables():
    db.connect()
    Project.create_table(True)
```

Our database queries change to ORM code - example from the docs:

```
for tweet in Tweet.select().where(Tweet.user == user, Tweet.is_published == True):
    print(tweet.user.username, '->', tweet.message)
```

hardbyte:
This is partially done - sqlalchemy and alembic are used to create and migrate the database. However we are not using the ORM in the main codebase.

PR #603, opened 2021-02-10
Use alembic to create database tables

### LINKED ISSUE #66: DB migration tool
Think about using [Alembic](http://alembic.zzzcomputing.com/en/latest/) to migrate between database schemas.

This is especially great to have once we are deployed for **real** in more than one place.

Aha! Link: https://csiro.aha.io/features/ANONLINK-20

hardbyte:
Adds sqlalchemy models for all database tables, uses alembic for migrations including initial database setup.

Note this doesn't remove all the raw SQL queries, but makes a start. 
Closes #66 and start to address part of #133

One side effect is deployments to k8s get working schema migrations :-)


wilko77 (commented):
I'm not sure that I am the right person to review this. 
Looks nice. Neater than before. 
I am a bit worried that we now have `entityservice/database/models` and `entityservice/models`. That's not so clean...

hardbyte:
Yeah the double up of `models` modules isn't ideal. As you can tell I have avoided the large task of moving off directly using `psycopg2` to using sqlalchemy in the application code.

I think the top level one could mostly be replaced by using the sqlalchemy models and using `marshmallow` to create the orm objects as per [this example](https://marshmallow.readthedocs.io/en/stable/examples.html#quotes-api-flask-sqlalchemy). I just don't  have the time to take that on.

wilko77 (approved):
I created an issue for the models stuff. Fine for now.

PR #605, opened 2021-02-16
Limit the maximum number of candidate pairs

### LINKED ISSUE #595: Limit maximum number of edges
I propose adding a feature where the service has a configured limit on the maximum number of edges. This would prevent the server from dying a slow and horrible death of swapping to a halt.

Probably makes sense to have a globally set limit or limits.

As the solver runs fully in memory the deployed high memory worker's memory limit imposes an upper limit.
Say the max memory was 16GiB, and estimating that each candidate pair is 40 bytes (8 Byte score, 4 x 8 Byte indicies into dataset and records):
```
(16 gibibytes) / (40 bytes) =
429 496 730
```

In the case where a user is attempting to download raw similarity scores instead of using the solver the limit may be much higher, but it still makes sense to perhaps configure the system to avoid creating arbitrarily large outputs.

I think it could also be useful to override this limit, less than the global limit, per run.

How about these names and defaults?

- `SOLVER_MAX_CANDIDATE_PAIRS = 100M`
- `SIMILARITY_SCORES_MAX_ CANDIDATE_PAIRS = 500M` (Could instead limit the file size in GiB)

The backend task would discard the data and cause the run to fail if these limits are exceeded.

hardbyte:
Adds two global settings to protect the service from running out of memory due to excessive numbers of candidate pairs being processed.

Closes #595


backend/entityservice/tasks/comparing.py
```diff
@@ -334,16 +332,26 @@ def reindex_using_encoding_ids(recordarray, encoding_id_list):
...
 
-###### results into file into minio
+    with new_child_span('check-within-candidate-limits') as scope:
+        global_candidates_for_run = get_candidate_count_for_run(run_id)
+        scope.span.log_kv({'global candidate count for run': global_candidates_for_run})
+
+        if global_candidates_for_run is not None and global_candidates_for_run > Config.SIMILARITY_SCORES_MAX_CANDIDATE_PAIRS:
+            log.warning(f"This run has created more than the global limit of candidate pairs. Setting state to 'error'")
+            with DBConn() as conn:
+                update_run_mark_failure(conn, run_id)
```
wilko77, line 351: So the user will never know why the run failed?

backend/entityservice/tasks/solver.py
```diff
@@ -25,6 +26,11 @@ def solver_task(similarity_scores_filename, project_id, run_id, dataset_sizes, p
         # https://github.com/data61/anonlink/issues/271
         candidate_pairs = _merge_similarities([zip(similarity_scores, dset_is0, dset_is1, rec_is0, rec_is1)], k=None)
         log.info(f"Number of candidate pairs after deduplication: {len(candidate_pairs[0])}")
+        if len(candidate_pairs[0]) > config.SOLVER_MAX_CANDIDATE_PAIRS:
+            log.warning(f"Attempting to solve with more than the global limit of candidate pairs.")
+            with DBConn() as conn:
+                update_run_mark_failure(conn, run_id)
```
wilko77, line 32: same here. Has the run table the ability to store an error message, and could that be passed on to the user?

wilko77 (commented):
It would be nicer to not even start a run if it is likely to exceed the resources. However, I don't have a good idea on how to estimate the number of candidate pairs above the threshold...

What I am a bit concerned about with your solution is that the server quietly fails. What's gonna be the first thing a user will do when he sees that his run errored? Try again. 
Would it be possible to have some sort of state-info field in the run table that we can include in the output of the run status endpoint?

hardbyte:
> It would be nicer to not even start a run if it is likely to exceed the resources. However, I don't have a good idea on how to estimate the number of candidate pairs above the threshold...

Yeah that would be nice, a very simple additional protection that can be pre-computed is including a limit on the number of comparisons?

> What I am a bit concerned about with your solution is that the server quietly fails. What's gonna be the first thing a user will do when he sees that his run errored? Try again. Would it be possible to have some sort of state-info field in the run table that we can include in the output of the run status endpoint?

It is certainly possible to add a new column to store error details in the `run` table - something like this in the alembic model:

```
err_msg = Column(String, nullable=True)
```

IMO I think that should be tackled separately though.

wilko77 (approved):

PR #608, opened 2021-02-17
Bug fix - Minio python API now requires a "DeleteObject"

hardbyte:
This object store deletion was only triggered in benchmarks and not in tests.


wilko77 (approved):
Good catch!
So we don't run at least one test that's big enough to call `merge_files`?

hardbyte:
Yeah I guess not, also odd that it seemed sporadic. That's never great

PR #612, opened 2021-02-21
Update env var name in object store init job

hardbyte:
Fix for https://github.com/data61/charts/pull/1#issuecomment-782947833


hardbyte:
A similar issue impacted the init db job. Would show up on k8s, as the default database server is the docker compose container name `db`.

wilko77 (approved):

PR #616, opened 2021-02-22
Update k8s labels

hardbyte:
Applies recommendations from 
https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels and https://helm.sh/docs/chart_best_practices/labels/#standard-labels


wilko77 (approved):
looks good to me...
Why didn't you set the `app.kubernetes.io/part-of` label as suggested on those two web sites?

hardbyte:
I considered it but I couldn't see where helm provides the "top level" application to use as the value.

Also I noticed that the charts we depend on (redis/minio etc) don't seem to do it (although I didn't look very hard).

PR #619, opened 2021-02-23
Add optional pod annotations to init jobs

hardbyte:


deployment/entity-service/values.yaml
```diff
@@ -129,6 +129,10 @@ api:
         cpu: 250m
         memory: 256Mi
 
+    ## Annotations added to the database init job's pod.
+    # podAnnotations: {}
```
wilko77, line 133: I thought those brackets shouldn't be there in a yaml config.

wilko77 (approved):

deployment/entity-service/values.yaml
hardbyte: I see them fairly regularly to indicate whether the expected type is a map or a list.

PR #624, opened 2021-02-26
Feature return error message

### LINKED ISSUE #617: Store and expose error message for a run
As discussed in https://github.com/data61/anonlink-entity-service/pull/605#issuecomment-780283375 

> Would it be possible to have some sort of state-info field in the run table that we can include in the output of the run status endpoint?

Add a new column to store error details in the run table - something like this in the alembic model:

```
err_msg = Column(String, nullable=True)
```

wilko77:
We now return a reason why a run failed.

closes #617


hardbyte (approved):
Nice

Issue #632, opened 2021-03-16
Implement P-Sig key filtering


hardbyte:
This service doesn't provide an api for client side filter of p-sig blocks, rather all candidate blocks are uploaded and the server [joins by block name](https://github.com/data61/anonlink-entity-service/blob/1b4a89ba2f639971eb8cf4288c45955a443d9d9b/backend/entityservice/tasks/comparing.py#L182).

See the [blocklib tutorial](https://blocklib.readthedocs.io/en/latest/tutorial/tutorial_blocking.html#Generate-Final-Blocks)

A related issue is the Anonlink Entity Service limits block names to 64 characters, which imposes a fairly low limit on p-sig's bloom filter size if we continue to use an encoding of the bloomfilter as the blocking key.

PR #633, opened 2021-03-22
Hash provided block identifier before storing in DB

hardbyte:
Related to issue #632.

Block names were limited to 64 chars, for P-Sig we may very well want to use block ids that are based on bloom filters... a 64 char limit would impose a max bloom filter size of 512 bits (assuming a perfect encoding unlike json).

This small change increases the column size to 512 chars, although happy to increase further or remove the character limitation.

### DIFF
--- a/backend/entityservice/encoding_storage.py
+++ b/backend/entityservice/encoding_storage.py
@@ -1,3 +1,5 @@
+from hashlib import blake2b
+
 import math
 from collections import defaultdict
 from itertools import zip_longest
@@ -33,12 +35,12 @@ def stream_json_clksnblocks(f):
     }
 
     :param f: JSON file containing clksnblocks data.
-    :return: Generator of (entity_id, base64 encoding, list of blocks)
+    :return: Generator of (entity_id, base64 encoding, iterable of hashed block names)
     """
     # At some point the user may supply the entity id. For now we use the order of uploaded encodings.
     for i, obj in enumerate(ijson.items(f, 'clknblocks.item')):
         b64_encoding, *blocks = obj
-        yield i, deserialize_bytes(b64_encoding), blocks
+        yield i, deserialize_bytes(b64_encoding), map(hash_block_name, blocks)
 
 
 def convert_encodings_from_base64_to_binary(encodings: Iterator[Tuple[str, str, List[str]]]):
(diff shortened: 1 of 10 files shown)


wilko77 (commented):
512 bytes for a block? 
Say we have 15 Million entities with about 1.5 Million blocks. If we use 512 bytes for every block, than we have to insert 6GB worth of block IDs into the database. 
However, the server only cares about matching up same block IDs. It does not care about the particular format. The minimum amount of bits to fully describe 1.5 Million different blocks is 21 bits. That's only about 30 MB of data. 

Why don't we do the following:
We let the user upload whatever they want. Then, when the server writes the block information into the database, it will map the user provided block IDs to some fixed-sized hash values and uses that as the internal block IDs.  128 bits should be plenty to all but eliminate the chance of a collision. 
On the plus side we would have less load on the database, as there is less data and plain integers instead of strings. What do you think?

hardbyte:
Oh I like that proposal a lot. So we will keep the DB constraint at 64 bytes, but hash the provided block id during import.

As for implementation, hashed via hmac with a run specific key? Key generated by the service on run creation and stored in the DB run table? `hmac_instance.hexdigest()` to convert to string for storage in the existing 64 char `block_id` column?

backend/entityservice/encoding_storage.py
```diff
@@ -33,12 +35,12 @@ def stream_json_clksnblocks(f):
...
 
     :param f: JSON file containing clksnblocks data.
-    :return: Generator of (entity_id, base64 encoding, list of blocks)
+    :return: Generator of (entity_id, base64 encoding, iterable of hashed block names)
     """
     # At some point the user may supply the entity id. For now we use the order of uploaded encodings.
     for i, obj in enumerate(ijson.items(f, 'clknblocks.item')):
         b64_encoding, *blocks = obj
-        yield i, deserialize_bytes(b64_encoding), blocks
+        yield i, deserialize_bytes(b64_encoding), map(hash_block_name, blocks)
```
hardbyte, line 43: Obviously we'd need to pass the run secret to `hash_block_name`, but I wanted to check whether you think its appropriate to modify the data here - as we are parsing the json file. Or just keep this function doing deserialization?

wilko77 (commented):
I don't think a keyed hash function is necessary.
We see the server as a semi-trusted third party. Thus, if the block IDs, as generated by the data providers, is leaking info about the PII, then the providers shouldn't upload it like that in the first place. The server must not be allowed to see that.
All the privacy protection has to happen on the client side. (like with the CLKs)
That means, if we have "leaky" block IDs, then the clients have to do something about it. Either encrypt, or use a keyed hash function. And only upload 'save' block IDs.

For the server part, the proposed hashing had the only purpose to create fixed sized indices, thus allowing clients to upload arbitrary block IDs. You could even just use the python inbuilt `hash` function.

backend/entityservice/tasks/comparing.py
```diff
@@ -277,6 +277,7 @@ def compute_filter_similarity(package, project_id, run_id, threshold, encoding_s
     task_span = compute_filter_similarity.span
 
     def new_child_span(name, parent_scope=None):
+        log.debug(name)
```
wilko77, line 280: that's not a good debug message.

wilko77 (approved):
looks good now

hardbyte:
Sorry, I forgot about this PR.

Just rebased after adding another test and combined two loops into one (importing blocks was hashing the names then counting the sizes in two separate loops). If/once the CI passes I'll merge

PR #659, opened 2021-08-17
Update postgres version

hardbyte:


wilko77 (commented):
why the different postgres versions in the chart vs the docker compose file?

hardbyte:
Fair question and its more complex than it seems.

The helm version `10.9.2` comes from the bitnami chart [here](https://github.com/bitnami/charts/blob/master/bitnami/postgresql/Chart.yaml#L29), be default that helm chart includes a postgres image of version `11.13.0` [defined in the default values file](https://github.com/bitnami/charts/blob/master/bitnami/postgresql/values.yaml#L77) and also reported in the [Chart.yaml as the app version](https://github.com/bitnami/charts/blob/master/bitnami/postgresql/Chart.yaml#L4).

This PR updates the docker-compose version to the latest stable release of the postgres docker image `13.4` and I tested it locally pretty extensively to work. 

I wouldn't be as certain that using the newest image with the bitnami helm chart would be as solid so I simply updated our chart's postgres dependency to the latest stable release (of the chart). 

Perhaps keeping the docker-compose version the same - pinning it to `11.13` - would be a better strategy to avoid unexpected surprises?

wilko77:
I'd prefer we use the same version for both.

wilko77 (approved):

PR #660, opened 2021-08-19
de-duplicate similarity scores

wilko77:
I noticed that the similarity score results can contain duplicates. 
The reason is that if a candidate pair appears in more than one block, they will produce more than one similarity score.

This PR introduces a simple unique value iterator to remove the duplicates from the similarity scores stream. The stream is already sorted, so duplicates will appear one after the other.


hardbyte (approved):
👍🏼 

So nice not to have to step about the multi part upload of unknown size with minio.

hardbyte:
Closes https://github.com/data61/anonlink/issues/271

PR #668, opened 2021-08-19
update to current alpine 3.14.1

wilko77:
and update packages to work with it.


hardbyte (approved):
🤞🏼 everything works with Python3.9

PR #677, opened 2021-08-23
Update the docker image upload script

hardbyte:


wilko77 (approved):
Looks like someone did some dusting...

PR #683, opened 2021-08-30
Delete encodings on project delete

wilko77:
Problem I want to solve: 
Encodings are not deleted during project deletion. That's kind of bad, as the db will eventually overflow and stop working.

Proposition:
We add the data provider id to an encoding in the database with a delete on cascade. This means that the encodings are going to be deleted if the data provider is removed. Same mechanism is used for the `encodingblocks` and `blocks`.


backend/entityservice/database/models/models.py
```diff
@@ -13,6 +13,7 @@ class Encoding(Base):
 
     encoding_id = Column(BigInteger, primary_key=True)
     encoding = Column(LargeBinary, nullable=False)
+    dp = Column(ForeignKey('dataproviders.id', ondelete='CASCADE'))
```
hardbyte, line 16: Being `nullable` was intentional so that users could upgrade?

Assuming we follow this up with a minor release, lets add a note that only "new" projects data gets cleaned up.

hardbyte (approved):
💥

PR #686, opened 2021-09-01
delete upload object store files after db ingestion

wilko77:
As the title says, once we have ingested the data, we clean up on the object store.


hardbyte (approved):
🧹 Loving all this picking up after ourselves

PR #699, opened 2021-10-03
Updates redis and postgres charts

hardbyte:
Thought we should update the dependencies before releasing a new version of the chart.

These charts allow slightly more customization during deployment. I've manually tested.


wilko77 (commented):
ooo, new shiny things..

wilko77 (approved):

PR #701, opened 2021-10-04
Make progress update more robust

hardbyte:
Saving comparison progress in redis now uses a configurable timeout for connecting, and has a retry using tenacity

The redis updates are pipelined together so they either all succeed or all fail.


wilko77 (approved):
nice

PR #702, opened 2021-10-05
Tolerate failure in comparison task

hardbyte:
First version for comment.

### DIFF
--- a/backend/entityservice/tasks/comparing.py
+++ b/backend/entityservice/tasks/comparing.py
@@ -404,8 +404,7 @@ def compute_filter_similarity(package, project_id, run_id, threshold, encoding_s
 
             mc = connect_to_object_store()
             try:
-                mc.put_object(
-                    Config.MINIO_BUCKET, result_filename, merged_file_iter, merged_file_size)
+                mc.put_object(Config.MINIO_BUCKET, result_filename, merged_file_iter, merged_file_size)
             except minio.S3Error as err:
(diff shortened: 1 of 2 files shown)


backend/entityservice/tasks/comparing.py
```diff
@@ -360,49 +370,50 @@ def reindex_using_encoding_ids(recordarray, encoding_id_list):
...
+            result_filename = Config.SIMILARITY_SCORES_FILENAME_FMT.format(generate_code(12))
+            task_span.log_kv({"edges": num_results})
+            log.info("Writing {} intermediate results to file: {}".format(num_results, result_filename))
+
+            mc = connect_to_object_store()
+            try:
+                mc.put_object(Config.MINIO_BUCKET, result_filename, merged_file_iter, merged_file_size)
+            except minio.S3Error as err:
+                log.warning("Failed to store result in minio: {}".format(err))
+                raise
```
wilko77: do you think we should wrap that in a separate retry with tenacity? We've come so far already, what if minio is temporarily not available?

wilko77 (approved):
Looks good.

In progress.py, we could wrap a retry around all the functions that interact with redis.

backend/entityservice/tasks/comparing.py
hardbyte: Yeah that sounds sensible, will do

PR #703, opened 2021-10-07
fixing jsonschema

wilko77:
the 4.x releases of jsonschema don't work with the current code base.
For now, we fix jsonschema to last know working version 3.2.0.

### DIFF
--- a/base/requirements.txt
+++ b/base/requirements.txt
@@ -1,3 +1,4 @@
+jsonschema==3.2.0
 alembic==1.6.5
 anonlink==0.14.2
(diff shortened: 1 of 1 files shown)


hardbyte:
Oh sorry about that, I think I must have upgraded or let dependabot upgrade that.

hardbyte (approved):

wilko77:
I think because we never explicitly defined which version we want to use, it installed the newest available version with anonlink-client.

hardbyte:
Ah okay

PR #716, opened 2021-11-23
👷 Update base alpine version to v3.14

hardbyte:
I noted that building the base image wasn't actually reproducible despite our overzealous pinning of requirements... so to make life easier I suggest we continue to pin the FROM docker image but allow alpine to find the consistent set of packages within that.

I've manually built and ran e2e tests via docker-compose.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -272,6 +272,43 @@ stages:
         inputs:
           helmVersionToInstall: 'latest'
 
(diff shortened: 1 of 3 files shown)


wilko77 (approved):
I really don't like how alpine works. 
Maybe it is time to move on...

radamson:
@hardbyte any plans to get this merged in?

